PR #15, opened 2014-06-26
adding support for reading from stdin

adkatrit:
It would be great to be able to load triples directly from stdin, so I made this simple implementation allowing for usage like `grep 'john' 30kmoviedata.nt | ./cayley http -stdin=true`

I've yet to fully play around with all the features of Cayley, so I may be missing something, but this seemed right to do.  It will at least serve to get the conversation started with regards to stdin as a datasource. 
Thanks!
Oh, and I signed the CLA.


adkatrit:
This PR is pretty stale.  Any ideas for direction on a 20 months newer Cayley?   Happy to take a swing.

adkatrit:
Closing for posterity.   If we want really want this.  We can take another clean whack at it.

PR #62, opened 2014-07-10
implement the -cpuprofile flag

miku:
Noticed the `-prof` flag was there, but the profiling was missing.

Added profiling as described in [Profiling Go Programs](http://blog.golang.org/profiling-go-programs).

Example output for the load operation (with leveldb) of `30kmoviesdata.nt`:

cpu

----

I signed the [CLA](https://developers.google.com/open-source/cla/individual).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,6 +12,7 @@ Alexander Peters <info@alexanderpeters.de>
 Google Inc.
 Jay Graves <jaywgraves@gmail.com>
 Jeremy Jay <jeremy@pbnjay.com>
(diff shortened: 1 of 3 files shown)


dennwc:
Branch has a conflict now. It's likely that we will switch to cobra soon, so this change maybe easier to do later. Furthermore it assumes that if `http` command was invoked it should not listen/serve pprof, but cmd  creates custom mux, so current code will not work for this case at all.

It was a pretty long time since PR was posted, so I guess you may not be interested in modifying it anymore. So I'm closing it now, but feel free to create recreate PR against HEAD.

PR #80, opened 2014-07-24
fix doc typo when wrapping code block

villadora:
As the result, the format is missing the document


dennwc:
File changed since then. Fixed already. But thanks anyway.

PR #84, opened 2014-07-28
Optimize iterators

pbnjay:
Simple updates. 1) Always Close() an interator when replaced, and 2) Give TripleStore a chance to optimize things.

(I've signed CLA)


dennwc:
We should reconsider iterator optimization in the new code base. Closing this, since too many things changed.

Issue #123, opened 2014-08-18
Delete node and delete all its relationships


tyteen4a03:
It would be useful if there is a way of deleting a node without having to know all its relationship beforehand.

robertmeta:
This was recently closed in the related issue.  yey!

PR #136, opened 2014-08-21
Issue 123 : Delete node and delete all its relationships

kris-lab:
Please have a look on simple (5min) implementation for "delete all relations of node" #123. Unfortunately there is no way to delete `node` itself over API.

Please let me know if this PR is useful or is there something I could do more?


dennwc:
New UI will be used instead. Opened a separate issue for node delete API.

PR #151, opened 2014-09-01
Update flags name and fix link

yageek:


barakmich:
`--quads` is actually correct here. The .md change is worth a fix.

PR #161, opened 2014-09-22
fix broken markdown in GremlinAPI.md

neonstalwart:
in addition to the markdown, i think there might be some copy/paste errors in the documentation regarding `TagArray`.  unfortunately, i'd only be guessing how to make it right, so maybe someone else could take a look and fix it.


barakmich:
Adding this to a doc-fix PR, thanks!

PR #174, opened 2014-10-15
support or-equals in mql

neonstalwart:
@barakmich i think i have a general idea of what needs to happen here:
- right after where [`reverse`](https://github.com/google/cayley/blob/84375200183f13816ae02c88f0da3ac559909289/query/mql/build_iterator.go#L122-L125) is determined i need to use a similar pattern - determine if the suffix is `"|="`, set a flag, strip the suffix from `pred`
- then somewhere a few lines later (i'm not exactly certain where) if the flag i set indicates an or-equals then i need to have a block of code something like what's [already in gremlin](https://github.com/google/cayley/blob/84375200183f13816ae02c88f0da3ac559909289/query/gremlin/build_iterator.go#L259-L270) to use `iterator.NewOr()`

the details are probably what will take me some time so i'll take any further advice you can give.

### DIFF
--- a/docs/MQL.md
+++ b/docs/MQL.md
@@ -84,3 +84,30 @@ A some_predicate C .
 exist.
 
 This combines with the reversal rule to create paths like ``"@a:!some_predicate"``
+
+## Union Queries (aka or-equals)
+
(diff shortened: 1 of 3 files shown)


dennwc:
Needs a heavy rewrite/rebase. Closing.

Issue #130, opened 2014-08-21
mongo empty result


changeYi:
whatever query, cayley return empty result for me.

eg:
(1) g.V("Humphrey Bogart").All()
return 
{"result": [ { "id": ""  } ]}

(2) g.V().GetLimit(2)
{ "result": [
  {  "id": "" },
  { "id": "" }
 ]}

this is my command
1, cayley_0.3.1_linux_amd64
2, config to mongodb
3, load data like : ./cayley load --config=cayley.cfg --triples=30kmoviedata.nq.gz
4, found 471705 document in mongo

dennwc:
Should be fixed with new versions of Mongo+Cayley

PR #180, opened 2014-11-02
Prevent empty values in cache (using mongo)

bkendall:
(This is a second PR for a subset of a previous PR that failed on travis because of the mongo versions - see below).

This PR prevents empty values from nodes that are not found from being added to the cache when the mongo backend is used. I was seeing almost exactly what issue #130 was describing and decided to sit down and debug my way through it. Removing the cache logic from the NameOf function pointed me in the right direction. The logic previously was adding to the cache even if the node was not found, which was then providing the empty strings I was seeing during the queries. It will now prevent a not found node from being added to the cache.

I tested these locally against mongo version `2.6.5` and the tests passed (`go test -v -backend=mongo ./...`)! My previous PR (#179) has the commit for adding them to travis, but since travis uses `2.4.x`, the tests failed (and I verified that they failed locally with a `2.4.x` mongo as well).

I have signed the CLA, as requested.

### DIFF
--- a/cayley_test.go
+++ b/cayley_test.go
@@ -392,6 +392,9 @@ func prepare(t testing.TB) {
 		}
 	case "mongo":
 		cfg.DatabasePath = "localhost:27017"
+		cfg.DatabaseOptions = map[string]interface{}{
+			"database_name": "cayley_test", // provide a default test database
+		}
 	default:
 		t.Fatalf("Untestable backend store %s", *backend)
 	}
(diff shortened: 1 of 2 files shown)


jaytula:
Just ran into this problem.  The PR works for me.  Thanks!

anandkumarpatel:
+1 ran into this issue, can this be merged

barakmich:
Cool. Yeah, that's a simple enough fix. 

Someday doing the Travis bit might make sense!

PR #192, opened 2014-12-23
Prevents repl panic

bcleenders:
Fixes https://github.com/google/cayley/issues/191

I have agreed to the CLA

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@
 # Please keep the list sorted.
 
 Alexander Peters <info@alexanderpeters.de>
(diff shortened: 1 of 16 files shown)


bcleenders:
So, how about now?

BTW: the reason there are so many commits, is because I did a git rebase with the master branch. Then it shows all commits twice.

kortschak:
LGTM

@barakmich can deal with history aesthetics.

barakmich:
Eh, we'll let the history be. It's a little messy, but it's following it's appropriate branch. (I've never been one who really likes squash commits, but it would be nice if it was linear from rebase point. So it goes.)

PR #194, opened 2015-01-03
Fix naming style issues found by lint

varadharajan:
Few naming issues were figured out by lint. This fixes them. Also the constants "quadsFirstFinal" and "quadsError" are not used anywhere. Do we need them?


varadharajan:
This PR is ready for review. Thanks

kortschak:
Not LGTM

These files are both generated code from ragel and should not be edited directly. The code is not for human consumption.

varadharajan:
Oops, I should've guessed it when i was wondering about the comments with line numbers. I'm closing this PR

PR #195, opened 2015-01-05
Allocate 82 bytes instead of 62 bytes in LevelDB's CreateKeyFor method

varadharajan:
In LevelDB's createKeyFor method we allocate size for 3 hashes but append 4 hashes. Eventhough append handles it intelligently, the code is misleading.

CLA already signed, PR ready for review


barakmich:
LGTM. That's a nice easy fix.

PR #199, opened 2015-01-14
Update GremlinAPI.md

ben-horner:
Graph.Vertex() was documented as returning a Query object, but that didn't make sense, and then in the Path documentation it said that .Vertex() returns a Path, which does make sense.


barakmich:
Fix is to actually mention that Query object is a subclass of Path object. (In a forthcoming PR)

PR #202, opened 2015-01-26
Ensure callers get errors when they happen

kortschak:
Previously we silently dropped portions or all of a block when a duplicate quad is found. We still fail now, but we tell someone.

Fixes #201.

PR #203, opened 2015-01-26
Make graph.Delta zero state invalid for use

kortschak:
Previously, an incorrectly initialised Delta (omission of Action) would
result in an Add operation. Make that detectable and return an error.


barakmich:
Merging this with #202.

PR #206, opened 2015-02-06
bench: Add some grunty Not queries to the integration test (derived from...

barakmich:
... existing queries)

### DIFF
--- a/cayley_test.go
+++ b/cayley_test.go
@@ -127,6 +127,41 @@ var benchmarkQueries = []struct {
 		},
 	},
 
+	// Exercises Not().Contains(), as above.
+	{
+		message: "the helpless checker, negated (films without Ingrid Bergman)",
+		long:    true,
+		query: `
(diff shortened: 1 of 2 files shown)


barakmich:
Merging on my own motion (docs and tests for a recent merge, based on previous tests).

kortschak:
I was trying to understand "the helpless checker, negated (films without Ingrid Bergman)", Can you explain that?

barakmich:
Sure. It's a hard query that produces nothing. The usual query is "people who starred in films, where film name is Casablanca"

The updated query is "people who starred in films, which do not star Ingrid Bergman, named Casablanca" -- which is the null set.

Compare to "people, not named Ingrid Bergman, who starred in a film, named Casablanca"

barakmich:
The "helpless checker" part is that we provide, at the very end, some context ("named Casablanca") -- without that clause, it's a very very broad query. This means the optimizer is going by Check() and not Next()

kortschak:
Ahh, github presents the line in my browser exacly so I expect no further text on the query - I missed the `.Is("Casablanca").All()`, which makes all the difference.

Issue #201, opened 2015-01-26
Bug: RDF with duplicated lines loading fails


LAlbertalli:
Hello

I'm experimenting with Cayley to import and use the Freebase dump. After the first import of Freebase I've noticed a lot of error regarding duplicated lines but from my understanding of Cayley code this shouldn't have affected the import.
Nevertheless I noticed the imported DB missed many nodes of the graph.
After more testing with smaller datasets I discovered that Cayley seems to stop importing after reaching a duplicated line.

To better test it I've created two sub dataset with the entity "http://rdf.freebase.com/ns/m.03j24kf" (Sir Paul McCartney) using a simple zgrep from the Freebase RDF dump of 12/18/2014. This dump (paul.txt) has line 1726 and 1759 duplicated. I've made also a dump without the duplication(paul1.txt). Both dump are available at https://github.com/LAlbertalli/cayley-examples/blob/master/cayley_dup.tgz

### Step to reproduce the issue:
1. Start Cayley with this command: `./cayley http -db=memstore -dbpath=$HOME/tests/entities/paul.txt -v=2 -alsologtostderr`
2. Run the query: `g.V("http://rdf.freebase.com/ns/m.01j7vsv").In("http://rdf.freebase.com/ns/music.artist.album").Out().All()`
3. Run the query: `g.V("http://rdf.freebase.com/ns/m.0rr6k3").In("http://rdf.freebase.com/ns/music.artist.track").Out().All()`
4. Restart Cayley: `./cayley http -db=memstore -dbpath=$HOME/tests/entities/paul1.txt -v=2 -alsologtostderr`
5. Run the query: `g.V("http://rdf.freebase.com/ns/m.01j7vsv").In("http://rdf.freebase.com/ns/music.artist.album").Out().All()`
6. Run the query: `g.V("http://rdf.freebase.com/ns/m.0rr6k3").In("http://rdf.freebase.com/ns/music.artist.track").Out().All()`

### Results:
* Query at line 3 return no result.

### Expected results:
* All the queries should return the same result (a list of 4382 element).

I've reproduced the issue both using memstore and Bolt as a backend (in production we plan to use Bolt) and using both the http interface and the repl.

### 1-3 Log:
````
I0125 19:24:22.686100 00356 cayley.go:108] Couldn't find a config file in either $CAYLEY_CFG or /etc/cayley.cfg. Going by flag defaults only.
I0125 19:24:22.688872 00356 cayley.go:170] Setting GOMAXPROCS to 4
[...]

PR #207, opened 2015-02-07
Add options to ignore duplicate or missing quad

LAlbertalli:
Add command line option to ignore duplicated quad when loading and ignore missing quads when deleting (two different options) as discussed in issue #201

The command line options are:
* -noerrdup to ignore duplicate
* -noerrdel to ignore missing

This is my first contribution to a go project so forgive me if some choice are not in line with the go coding style

Also note that leveldb doesn't actually detect duplicated line in the same block because is loaded in the db with the same batch. I don't know the impact of such bug.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -123,3 +123,23 @@ Optionally disable syncing to disk per transaction. Nosync being true means much
   * Default: "cayley"
 
 The name of the database within MongoDB to connect to. Manages its own collections and indicies therein.
+
+## Per-Replication Options
+
+The `replication_options` object in the main configuration file contains any of these following options that change the behavior of the replication manager.
+
+### All
+
+#### **`ignore_missing`**
+
+  * Type: Boolean
+  * Default: false
+
+Optionally ignore missing quad on delete.
+
+#### **`ignore_duplicate`**
+
+  * Type: Boolean
+  * Default: false
+
+Optionally ignore duplicated quad on add.
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -43,6 +43,7 @@ var (
 	}
 	hashSize         = sha1.Size
 	localFillPercent = 0.7
+
 )
 
 type Token struct {
@@ -208,6 +209,12 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta) error {
 		for _, d := range deltas {
 			err := qs.buildQuadWrite(tx, d.Quad, d.ID.Int(), d.Action == graph.Add)
 			if err != nil {
+				if err == graph.ErrQuadExists && *graph.NoErrorDup{
+					continue
+				}
+				if err == graph.ErrQuadNotExist && *graph.NoErrorDel{
+					continue
+				}
 				return err
 			}
 			delta := int64(1)
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -185,7 +185,7 @@ var (
 	metaBucket = []byte("meta")
 )
 
-func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDup, ignoreMiss bool) error {
+func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOpts) error {
 	oldSize := qs.size
 	oldHorizon := qs.horizon
 	err := qs.db.Update(func(tx *bolt.Tx) error {
@@ -209,10 +209,10 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDup, ignoreMiss boo
 		for _, d := range deltas {
 			err := qs.buildQuadWrite(tx, d.Quad, d.ID.Int(), d.Action == graph.Add)
 			if err != nil {
-				if err == graph.ErrQuadExists && ignoreDup{
+				if err == graph.ErrQuadExists && ignoreOpts.IgnoreDup{
 					continue
 				}
-				if err == graph.ErrQuadNotExist && ignoreMiss{
+				if err == graph.ErrQuadNotExist && ignoreOpts.IgnoreMissing{
 					continue
 				}
 				return err
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -185,7 +185,7 @@ var (
 	metaBucket = []byte("meta")
 )
 
-func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta) error {
+func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDuplicate bool, ignoreMissing bool) error {
 	oldSize := qs.size
 	oldHorizon := qs.horizon
 	err := qs.db.Update(func(tx *bolt.Tx) error {
@@ -209,10 +209,10 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta) error {
 		for _, d := range deltas {
 			err := qs.buildQuadWrite(tx, d.Quad, d.ID.Int(), d.Action == graph.Add)
 			if err != nil {
-				if err == graph.ErrQuadExists && *graph.NoErrorDup{
+				if err == graph.ErrQuadExists && ignoreDuplicate{
 					continue
 				}
-				if err == graph.ErrQuadNotExist && *graph.NoErrorDel{
+				if err == graph.ErrQuadNotExist && ignoreMissing{
 					continue
 				}
 				return err
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -185,7 +185,7 @@ var (
 	metaBucket = []byte("meta")
 )
 
-func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDuplicate bool, ignoreMissing bool) error {
+func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDup, ignoreMiss bool) error {
 	oldSize := qs.size
 	oldHorizon := qs.horizon
 	err := qs.db.Update(func(tx *bolt.Tx) error {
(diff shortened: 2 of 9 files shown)


barakmich:
So I think you're about halfway there.

The flags you're adding are really configuration bits for a QuadWriter (of which there is now only single) and some subset of these may be piped into AddDelta() as another argument (ie, a new DeltaOpts struct). So you did flags correct-ish but really you'd want this piped through the single-point-of-flags which is in the `/config` directory.

Yeah, it's a little onerous. On the other hand, you can write the plumbing which will be used for other writer configuration options (replication URLs, etc) in the future, which is plumbing that's *really* useful.

When that's there, I think you're close. The `delete` is often correct, and I'd have to revisit the add logic, but I'm happy to allow these ignores as a general idea.

kortschak:
See http://commandcenter.blogspot.com.au/2014/01/self-referential-functions-and-design.html and http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis for discussions about how we can deal with options in a nicer way.

barakmich:
@kortschak That is really cool. 

There would still need to be some layer by which the configuration gets parsed from a file, but passing them around suddenly becomes really interesting (and perfectly cromulent to this PR). The configuration parse too, for that matter. (UnmarshalJSON a dict into a list of options appropriate for the package). The trick being to not import the package into the config file parser, but that's not too bad either (an options.go for the appropriate package)

The real win comes when you start wanting to make calls into packages separably, ie, like an API. Beyond that there's not much difference between `(<blah>, opts... Option)` and `(<blah>, opt *Options)`. The latter is forced to be crufty (`(<blah>, nil)` in the best case), the former cleaner.

So to what extent is that the case here? I think the answer is "somewhat" --useful as it relates to QuadWriters, maybe less as it applies to AddDelta(). A key difference seems to be that one is an object that manages some long-lived state, the other mostly functional in its way (that is, the options (in this case, ignore opts) are ephemeral to the single call). Example: setting IgnoreNoExist on the QuadWriter as a functional opt as suggested, and then passing it as a struct opt (based on current QW state) to AddDelta() makes a lot of sense, API wise.

That's about where I'd draw the boundary. Open for further thoughts!

kortschak:
Yes, I was thinking of this as the underlying API, and a map[string]Option that is used in the option parsing (still thinking in terms of the Go API which I will look at over the next few days - thank you for that).

I think it can go in later if this PR deals with the options via the Config struct as other options are.

barakmich:
Agreed. @LAlbertalli -- if you want to continue just using a Config struct as I suggested, we'll revisit when the time comes. :+1:

LAlbertalli:
@barakmich @kortschak Thanks for the suggestions and the critics.

For using the config, you are right, actually I'm using it for another PR I'm working on. So no problem to add it. My first idea was to add the flag in the QuadWriter but I simply did not understood how it works (I'm a GO rookie) and as a configuration option for the QuadStore was easier.

So, just to be clear, it is ok if I just rework the code to use also the config file and to got through the config code?

Finally, I still have doubt of the behavior of LevelDB. To me it is a bug but not sure if the effort to fix it is worthy.

Thanks
L.

barakmich:
>So, just to be clear, it is ok if I just rework the code to use also the config file and to got through the config code?

Yep, sounds good to me.

> Finally, I still have doubt of the behavior of LevelDB. To me it is a bug but not sure if the effort to fix it is worthy.

What's the bug? It may warrant another Issue or PR.

LAlbertalli:
@barakmich 
Sorry, I wasn't probably clear in my PR.
Level DB doesn't always identifies duplicates. The issue is with batch loading. If there's a duplicate entry in a quadset, it is not identified because LevelDB does batch writes so when the code does a get to load find if a record exists, it will not find the duplicated quad. I think the solution would be to keep a map of batch already loaded but not yet written but I'm not totally sure if it worths the efforts. My main concerns right now is if it could cause inconsistencies...

LAlbertalli:
I'm working on implementing the configurations as suggested but I'm facing some problems:
* I don't want to change the QuadStore interface so the best way is to put the options in the DbOptions field but:
   * From my understanding of GO, I have to replicate the code in all implementations but it is not a good idea for code maintainability
   * Also it will be difficult to move the option to the QuadWriter (for compatibility)
* I want to keep the runtime flag for the config because it makes sense in my opinion (e.g: we want to ignore errors in massive load but not in normal operations, and changing the config file makes little sense).

To solves partially these issues I have this in mind but not sure if it could be a good idea:
* Add the Options struct as an anonymous field to all the QuadStore structs in order to access them through composition
* Add specific methods to Options that access the right options and, if missing, read them from the flag and then return the default
This makes sense? I know it is a big change on how the Quaestores are implemented but to me make sense.

kortschak:
You have the options in the respective newQuadStore functions, where the types are concrete, so you can set them there.  I would change the names, s/NoErrorDup/allowDup/g and s/NoErrorDel/ignoreMissing/g - these are now fields rather than globals and so can be unexported.

kortschak:
Actually, s/NoErrorDup/ignoreDup/g.

barakmich:
Yep. NewQuadWriter already takes a opts Options, so that's where the option from config should go generally. Feel free to add a field like this to the config-- I think "writer_opts" is probably closest.

Then, instead of holding these options in each QuadStore, pass them as arguments to the one function that QuadWriter calls: ApplyDeltas(). So that signature alone (for each QuadStore) will change to accept a new WriteOptions struct. And then based on that, you can make similar checks as the flag you'd created.

And when we get around to functional-options the NewQuadWriter business will be much cleaner.

LAlbertalli:
Ok, I have implemented all the changes we discussed.
Right now the configuration file has priority over the the flag passed to the db at startup, not sure if it is a desired behavior but I kept it.

graph/bolt/quadstore.go
```diff
@@ -184,7 +185,7 @@ var (
 	metaBucket = []byte("meta")
 )
 
-func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta) error {
+func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDuplicate bool, ignoreMissing bool) error {
```
kortschak: Delete first bool type. Also below.

The length of this signature troubles me for such a simple operation.

kortschak:
Flags should take precedence over configs.

LAlbertalli:
Shortened the function signatures and changed the priority of flags over configuration as suggested by @kortschak .
One note, right now, if I'm not wrong, the config file has priority over the flags.

docs/Configuration.md
```diff
@@ -123,3 +123,23 @@ Optionally disable syncing to disk per transaction. Nosync being true means much
   * Default: "cayley"
 
 The name of the database within MongoDB to connect to. Manages its own collections and indicies therein.
+
+## Per-Replication Options
```
barakmich: Right notion. I even would have called it `replication_options` myself before. But I'm open to other suggestions, now that I see it in writing

(Thank you for keeping the docs in sync, btw!)

graph/bolt/quadstore.go
```diff
@@ -184,7 +185,7 @@ var (
 	metaBucket = []byte("meta")
 )
 
-func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta) error {
+func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreDup, ignoreMiss bool) error {
```
barakmich: I'll take it. I might have done a struct for this case, but (at the moment) it's two bools, and may get updated. So, this may get upgraded to full options or to a struct if we add another flag, but for now, I'm willing to go with it.

graph/iterator/mock_ts_test.go
```diff
@@ -35,7 +35,7 @@ func (qs *store) ValueOf(s string) graph.Value {
 	return nil
 }
 
-func (qs *store) ApplyDeltas([]graph.Delta) error { return nil }
+func (qs *store) ApplyDeltas([]graph.Delta, bool, bool) error { return nil }
```
barakmich: I'll take it. I might have done a struct for this case, but (at the moment) it's two bools, and may get updated. So, this may get upgraded to full options or to a struct if we add another flag, but for now, I'm willing to go with it.

graph/leveldb/quadstore.go
```diff
@@ -250,12 +257,17 @@ func (qs *QuadStore) buildQuadWrite(batch *leveldb.Batch, q quad.Quad, id int64,
 	} else {
 		entry.Quad = q
 	}
-	entry.History = append(entry.History, id)
 
-	if isAdd && len(entry.History)%2 == 0 {
-		glog.Error("Entry History is out of sync for", entry)
-		return errors.New("odd index history")
+	if isAdd && len(entry.History)%2 == 1 {
```
barakmich: Yeah, this is clearer. :)

graph/quadstore.go
```diff
@@ -44,7 +44,7 @@ type Value interface{}
 type QuadStore interface {
 	// The only way in is through building a transaction, which
 	// is done by a replication strategy.
-	ApplyDeltas([]Delta) error
+	ApplyDeltas([]Delta, bool, bool) error
```
barakmich: I'll take it. I might have done a struct for this case, but (at the moment) it's two bools, and may get updated. So, this may get upgraded to full options or to a struct if we add another flag, but for now, I'm willing to go with it.

barakmich:
Yep, looks pretty good. Changed my mind mid-review. quadwriter.go should have a `type IgnoreOpts struct` with your two booleans. This then gets passed to AddDeltas (and is a member of Single), which cleans things up a little.
 
We'll probably need it anyway (to store the options) if they're passed in other ways. 

Other than that, flag location looks alright, add logic is fine, and the rest looks good.

Rename `replication_options` in `config/config.go` if you like. I can see I had it there already, but l'm not sure I like the name anymore. What makes sense to you?

LAlbertalli:
Done the change with the struct.
I haven't changed the `replication_options`.

barakmich:
Awesome. Let's do this! LGTM

steelbrain:
:+1:  Awesome! The cayley logs will be much much cleaner with this applied.

Issue #64, opened 2014-07-12
Higher-level Go API


barakmich:
Tracking for this discussion that keeps happening -- now it's got a number.

Basically, it'd be great to write queries (or, more explicitly, build iterator trees) without having to be so mechanistic in building the iterator trees. Exposing such an API makes it easier for people using Cayley as a library, makes it easier to write query languages (that have repeated patterns in a lot of places) and, perhaps, make it easier to talk about external API connections -- essentially, the wire format for building iterator trees. Which would allow interesting bindings for other languages without catting together Javascript. 

This may be directly related to the Gremlin API, in that one may influence the other and vice versa.

flosse:
I hope that's the right place to ask:
Where can I find some introductions for using cayley as a library? Is there a hello world example? And do you know a real world example of a go application that's using cayley (no matter which API is used) ?

iddan:
Documentation for using Cayley as a lib can be found here: https://github.com/cayleygraph/cayley/blob/master/docs/Quickstart-As-Lib.md.
As far as I understand this feature was resolved and can be closed. @robertmeta @dennwc

dennwc:
@iddan Yes, I think we can close it. Most of it is implemented I believe. For everything else we have a more focused issues open.

Issue #121, opened 2014-08-15
Cursors and Pagination


ghost:
Is there a way to offset a query, e.g:  g.V().GetLimit(limit, offset)? Such a feature would be very handy for pagination purposes.

fisnik:
Any news on this feature - i think this is a must as without it, it would be hard to considering Cayley for production on anything dependent on pagination. 

Also, how would the cursors be handled? Would the presence of cursors become an impediment on a live production deployment? If you have thousands of clients requesting paginated data, than you would have as many cursors as clients.

jzelinskie:
Totally stealing @barakmich's words:

You can serialize the iterators and encode them and pass them as a token to the client. Thus, the client can just pass the token back and cayley deserializes it and continues iterating for results.

iddan:
I think there should be a streamable content type for Gizmo to allow

dennwc:
@iddan This would require some interactive API like WebSockets or gRPC, but it's easier to implement than serializing the iterator chain (although achievable).

iddan:
> This would require some interactive API like WebSockets or gRPC
I think writing multiple chunks to the HTTP message will be good enough. In Go's `http` I think you can use `Response.Write()` and `Response.Flush()`

PR #208, opened 2015-02-09
Go API

barakmich:
First run at a Go-importable API for Cayley. This would be the first step into using Cayley as a library and tackling #64 to a degree.

It's more or less a port of Gremlin into Go syntax with a little magic behind the scenes. Evaluation still remains unchanged (that'd be a nice followup PR -- get all top level nodes, get all tags, etc) but it's much, much easier to build iterator trees now.

Needs a large amount of documentation before committing. Sending the PR for initial feedback.

/cc @sdboyer @kortschak

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -166,7 +166,7 @@ type Description struct {
 }
 
 // A curried function that can generates a new iterator based on some prior iterator.
-type MorphismFunc func(Iterator) Iterator
+type ApplyMorphismFunc func(QuadStore, Iterator) Iterator
 
 type Nexter interface {
 	// Next advances the iterator to the next value, which will then be available through
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -165,6 +165,9 @@ type Description struct {
 	Iterators []Description  `json:",omitempty"`
 }
 
+// A curried function that can generates a new iterator based on some prior iterator.
+type MorphismFunc func(Iterator) Iterator
+
 type Nexter interface {
 	// Next advances the iterator to the next value, which will then be available through
 	// the Result method. It returns false if no further advancement is possible, or if an
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -165,8 +165,8 @@ type Description struct {
 	Iterators []Description  `json:",omitempty"`
 }
 
-// A curried function that can generates a new iterator based on some prior iterator.
-type ApplyMorphismFunc func(QuadStore, Iterator) Iterator
+// ApplyMorphism is a curried function that can generates a new iterator based on some prior iterator.
+type ApplyMorphism func(QuadStore, Iterator) Iterator
 
 type Nexter interface {
 	// Next advances the iterator to the next value, which will then be available through
--- a/graph/path/path.go
+++ b/graph/path/path.go
@@ -32,6 +32,9 @@ type Path struct {
 }
 
 func StartPath(qs graph.QuadStore, nodes ...string) *Path {
+	if len(nodes) == 0 {
+		return PathFromIterator(qs, qs.NodesAllIterator())
+	}
 	return &Path{
 		stack: []morphism{
 			isMorphism(qs, nodes...),
@@ -43,7 +46,7 @@ func StartPath(qs graph.QuadStore, nodes ...string) *Path {
 func PathFromIterator(qs graph.QuadStore, it graph.Iterator) *Path {
 	return &Path{
 		stack: []morphism{
-			intersectIteratorMorphism(it),
+			iteratorMorphism(it),
 		},
 		qs: qs,
 	}
@@ -61,6 +64,11 @@ func (p *Path) Reverse() *Path {
 		newPath.stack = append(newPath.stack, p.stack[i].Reversal())
 	}
 	return newPath
+}
+
+func (p *Path) Is(nodes ...string) *Path {
+	p.stack = append(p.stack, isMorphism(p.qs, nodes...))
+	return p
 }
 
 func (p *Path) Tag(tags ...string) *Path {
@@ -84,6 +92,16 @@ func (p *Path) And(path *Path) *Path {
 
 func (p *Path) Or(path *Path) *Path {
 	p.stack = append(p.stack, orMorphism(path))
+	return p
+}
+
+func (p *Path) Follow(path *Path) *Path {
+	p.stack = append(p.stack, followMorphism(path))
+	return p
+}
+
+func (p *Path) FollowReverse(path *Path) *Path {
+	p.stack = append(p.stack, followMorphism(path.Reverse()))
 	return p
 }
 
@@ -149,10 +167,10 @@ func inMorphism(qs graph.QuadStore, via ...interface{}) morphism {
 	}
 }
 
-func intersectIteratorMorphism(it graph.Iterator) morphism {
+func iteratorMorphism(it graph.Iterator) morphism {
 	return morphism{
 		"iterator",
-		func() morphism { return intersectIteratorMorphism(it) },
+		func() morphism { return iteratorMorphism(it) },
 		func(subIt graph.Iterator) graph.Iterator {
 			and := iterator.NewAnd()
 			and.AddSubIterator(it)
@@ -186,6 +204,17 @@ func orMorphism(path *Path) morphism {
 			and.AddSubIterator(it)
 			and.AddSubIterator(subIt)
 			return and
+		},
+	}
+}
+
+func followMorphism(path *Path) morphism {
+	return morphism{
+		"follow",
+		func() morphism { return followMorphism(path.Reverse()) },
+		func(base graph.Iterator) graph.Iterator {
+			p := path.MorphismFunc()
+			return p(base)
 		},
 	}
 }
--- a/graph/path/path.go
+++ b/graph/path/path.go
@@ -23,21 +23,24 @@ import (
 type morphism struct {
 	Name     string
 	Reversal func() morphism
-	Apply    graph.MorphismFunc
+	Apply    graph.ApplyMorphismFunc
 }
 
 type Path struct {
 	stack []morphism
-	qs    graph.QuadStore
+	qs    graph.QuadStore // Optionally. A nil qs is equivalent to a morphism.
+}
+
+func (p *Path) IsMorphism() bool { return p.qs == nil }
+
+func StartMorphism(nodes ...string) *Path {
+	return StartPath(nil, nodes...)
 }
 
 func StartPath(qs graph.QuadStore, nodes ...string) *Path {
-	if len(nodes) == 0 {
-		return PathFromIterator(qs, qs.NodesAllIterator())
-	}
 	return &Path{
 		stack: []morphism{
-			isMorphism(qs, nodes...),
+			isMorphism(nodes...),
 		},
 		qs: qs,
 	}
@@ -67,7 +70,7 @@ func (p *Path) Reverse() *Path {
 }
 
 func (p *Path) Is(nodes ...string) *Path {
-	p.stack = append(p.stack, isMorphism(p.qs, nodes...))
+	p.stack = append(p.stack, isMorphism(nodes...))
 	return p
 }
 
@@ -77,11 +80,11 @@ func (p *Path) Tag(tags ...string) *Path {
 }
 
 func (p *Path) Out(via ...interface{}) *Path {
-	p.stack = append(p.stack, outMorphism(p.qs, via...))
+	p.stack = append(p.stack, outMorphism(via...))
 	return p
 }
 func (p *Path) In(via ...interface{}) *Path {
-	p.stack = append(p.stack, inMorphism(p.qs, via...))
+	p.stack = append(p.stack, inMorphism(via...))
 	return p
 }
 
@@ -92,6 +95,11 @@ func (p *Path) And(path *Path) *Path {
 
 func (p *Path) Or(path *Path) *Path {
 	p.stack = append(p.stack, orMorphism(path))
+	return p
+}
+
+func (p *Path) Except(path *Path) *Path {
+	p.stack = append(p.stack, exceptMorphism(path))
 	return p
 }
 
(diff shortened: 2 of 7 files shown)


barakmich:
FWIW, a quick run at a separate app using Cayley as a library:
```go
package main

import (
	"fmt"

	_ "github.com/google/cayley/graph/memstore"
	"github.com/google/cayley/graph/path"
	_ "github.com/google/cayley/writer"
)

func main() {
	// Could have a helper function for memstore like cases
	qs, _ := graph.NewQuadStore("memstore", "", nil)
	w, _ := graph.NewQuadWriter("single", qs, nil)

	w.AddQuad(quad.Quad{"food", "is", "good", ""})

	p := path.StartPath(qs, "food").Out("is")

	// Could be added to the library
	it := p.BuildIterator()
	for graph.Next(it) {
		v := qs.NameOf(it.Result())
		fmt.Println(v)
	}
}

```

graph/iterator.go
```diff
@@ -153,6 +153,9 @@ type Description struct {
 	Iterators []Description  `json:",omitempty"`
 }
 
+// A curried function that can generates a new iterator based on some prior iterator.
+type ApplyMorphismFunc func(QuadStore, Iterator) Iterator
```
kortschak: s/ApplyMorphismFunc/ApplyMorphism/

Also per go vet, "// ApplyMorphism is a "...

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	return p.BuildIteratorOn(p.qs)
+}
+
+func (p *Path) BuildIteratorOn(qs graph.QuadStore) graph.Iterator {
+	f := p.MorphismFunc()
+	return f(qs, qs.NodesAllIterator())
+}
+
+func (p *Path) MorphismFunc() graph.ApplyMorphismFunc {
```
kortschak: s/Func//g

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (p *Path) BuildIterator() graph.Iterator {
+	if p.IsMorphism() {
+		panic("Building an iterator from a morphism. Bind a QuadStore with BuildIteratorOn(qs)")
+	}
+	return p.BuildIteratorOn(p.qs)
+}
+
+func (p *Path) BuildIteratorOn(qs graph.QuadStore) graph.Iterator {
+	f := p.MorphismFunc()
```
kortschak: s/Func//

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+func (p *Path) Out(via ...interface{}) *Path {
+	p.stack = append(p.stack, outMorphism(via...))
+	return p
+}
+func (p *Path) In(via ...interface{}) *Path {
+	p.stack = append(p.stack, inMorphism(via...))
+	return p
+}
+
+func (p *Path) And(path *Path) *Path {
```
kortschak: s/path/p/g

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+func followMorphism(path *Path) morphism {
+	return morphism{
+		"follow",
+		func() morphism { return followMorphism(path.Reverse()) },
+		func(qs graph.QuadStore, base graph.Iterator) graph.Iterator {
+			p := path.MorphismFunc()
+			return p(qs, base)
```
kortschak: `return p.MorphismFunc()(qs, base)`

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+	and.AddSubIterator(lto)
+	return iterator.NewHasA(viaPath.qs, and, out)
+}
+
+func buildViaPath(qs graph.QuadStore, via ...interface{}) *Path {
+	if len(via) == 0 {
+		return PathFromIterator(qs, qs.NodesAllIterator())
+	} else if len(via) == 1 {
+		v := via[0]
+		if path, ok := v.(*Path); ok {
```
kortschak:
```
switch v := v.(type) {
case *Path:
    return v
case string:
    return StartPath(qs, str)
default:
    panic("Invalid type passed to buildViaPath.")
}

File: graph/path/path.go [hunk 7]
```diff
@@ -0,0 +1,296 @@
...
+	"github.com/google/cayley/quad"
+)
+
+type morphism struct {
+	Name     string
+	Reversal func() morphism
+	Apply    graph.ApplyMorphismFunc
+}
+
+type Path struct {
```
kortschak: If you add an `err error` field and an `Err() error` method you can keep the fluent style without the need to resort to panics.

Then just add an `if p.err != nil { return }` to each method as a preamble.

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+	p.stack = append(p.stack, followMorphism(path))
+	return p
+}
+
+func (p *Path) FollowReverse(path *Path) *Path {
+	p.stack = append(p.stack, followMorphism(path.Reverse()))
+	return p
+}
+
+func (p *Path) BuildIterator() graph.Iterator {
```
kortschak: Return error rather than panic.

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+		in, out = out, in
+	}
+	lto := iterator.NewLinksTo(viaPath.qs, it, in)
+	and := iterator.NewAnd()
+	and.AddSubIterator(iterator.NewLinksTo(viaPath.qs, viaPath.BuildIterator(), quad.Predicate))
+	and.AddSubIterator(lto)
+	return iterator.NewHasA(viaPath.qs, and, out)
+}
+
+func buildViaPath(qs graph.QuadStore, via ...interface{}) *Path {
```
kortschak: Return an error instead of panic.

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+		} else {
+			panic("Invalid type passed to buildViaPath.")
+		}
+	}
+	var strings []string
+	for _, s := range via {
+		if str, ok := s.(string); ok {
+			strings = append(strings, str)
+		} else {
+			panic("Non-string type passed to long Via path")
```
kortschak: `return nil, errors,New("non-string type passed to long Via path")`

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+		}
+	}
+	return StartPath(qs, strings...)
```
kortschak: plus err

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+			return it
+		}}
+}
+
+func outMorphism(via ...interface{}) morphism {
+	return morphism{
+		"out",
+		func() morphism { return inMorphism(via...) },
+		func(qs graph.QuadStore, it graph.Iterator) graph.Iterator {
+			path := buildViaPath(qs, via...)
```
kortschak: `path, err := buildViaPath(qs, via...)`

graph/path/path.go
kortschak: If you add an `err error` field and an `Err() error` method you can keep the fluent style without the need to resort to panics.

Then just add an `if p.err != nil { return }` and a recover block to set the err field to each method as a preamble.

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+import (
+	"github.com/google/cayley/graph"
+	"github.com/google/cayley/graph/iterator"
+	"github.com/google/cayley/quad"
+)
+
+type morphism struct {
+	Name     string
+	Reversal func() morphism
+	Apply    graph.ApplyMorphismFunc
```
kortschak: s/Func//

graph/path/path.go
```diff
@@ -0,0 +1,296 @@
...
+		func(_ graph.QuadStore, subIt graph.Iterator) graph.Iterator {
+			and := iterator.NewAnd()
+			and.AddSubIterator(it)
+			and.AddSubIterator(subIt)
+			return and
+		},
+	}
+}
+
+func andMorphism(path *Path) morphism {
```
kortschak: s/path/p/g

kortschak:
LGTM with minor comments (apologies for spam in inbox - writing before thinking).

graph/path/path.go
kortschak: Actually, this probably is a programmer error, so panic is fine.

sdboyer:
been having some odd dizziness recently that has left me a little impaired, so, sorry for the delay... 

@kortschak 's notes look good to me, as does this approach in general.

one thought i had: incorporating write operations more directly with traversals. so, i traverse and grab a set of vertices, and with a single method call on that result, perform a write. this could take different shapes, of course, but generally of the form where the result set are taken being the subjects OR objects, and the method call gives the remainder of the quad.

probably good for a follow-up, though.

graph/path/path.go
barakmich: My thinking exactly. A program which executes such a query (namely, one that is malformed) should panic. The query may be *wrong* but at least it's *valid*. ;)

graph/path/path.go
barakmich: The only panics are programmer error -- passing wrong types, making invalid assumptions, etc. With a richer type system, this would be in the type checker, but panic is my solution to the madness.

kortschak: True. Good as is then. Obviously when docs go on, the acceptable types need to be defined.

barakmich:
@kortschak Fixed based on your comments

@sdboyer This would be a potentially cool thing. You now have a Path object. With it, you can create meta write operations, eg, WriteAsSubject(Path, string, string)

Still needs comments everywhere before I commit it.

And I've been thinking about an /api (/cayleyapi? /cayley?) folder, which may import things like Path and whatnot to create a nice set of hooks. The above example might be then (assuming a /cayley folder)

```go
qs, w := cayley.Memstore()
w.AddQuad(...)
p := cayley.StartPath()...
set := cayley.GetAllTags(p)
```
And so on. 

Paths have clear use within other query languages (ie, simplifying a lot of Gremlin, if we'd like) hence it being a separate package to itself.

kortschak:
There is a short talk on naming (including package naming) at http://talks.golang.org/2014/names.slide and blog post here http://blog.golang.org/package-names.

My preference would be that it get the root cayley and the command go in cayley/cmd/cayley, but I don't think that is what you want to do.

sdboyer:
@barakmich yeah, i really like the separation of the path concept. lot of value there.

also +1 to rolling up a separate package that's intended for direct import. beyond it simply being more convenient, it's the right way to provide a coherent, consolidated starting point for anyone wanting to directly import and work with Cayley.

no opinions on the appropriate name for it, though :)

sdboyer:
`caypi` :black_joker:

kortschak:
Another thing that comes out of this is that it will probably prompt a need to make the form of logging more configurable (rather than glog being imposed). I've had thoughts about this, but I don't see a nice or easy solution. Also, I still need to go through and tighten up the error handling and propagation (this latter would help the former).

andrew-d:
@barakmich Is this in a merge-able state?  I'd love to be able to use this - is there anything I can do to help get it merged?

barakmich:
@andrew-d Yes! It's pretty close -- what would really help is to help me write some comments based on the Gremlin API definitions, so that this piece is godoc-able.

What would also be nice (perhaps a separate PR) is the notion of a cursor -- a wrapper around an iterator which calls graph.Next() in various ways (each result, each top-level result, etc)

So if you're up for helping out, I can merge this now for Godocing or you can branch from this. I could definitely use the help, and it's a great way to start

kortschak:
The main concern I have here is that we don't yet deal with errors properly. This is OK when we are providing an engine that another tightly integrated system is built on that like we have here, and where we are free to change internals because there is no advertised public access point, but not so good when there is a defined Go API.

The issue is that errors are logged to glog in a number of places rather than being returned, and worse, in some places it is not at all clear how an error would be returned - this is the main reason I have lagged on this.

0. some graph operations are failable and easily fixed - e.g. graph.Iterator.Close should return error.
1. some graph stores have operations that may fail and are necessary for the iteration API - this is less easily fixed, e.g. the Next method in leveldb (less concerning) and mongo (more concerning).
2. the iteration API needs to remain simple (Next should return a bool only), so how to signal error states.

I am thinking that to keep the iteration API simple, we should add an Err() error method to graph.Iterator to check on false. This is potentially fraught (people won't use it), but it keeps the API use manageable.

I may be able to find some time in the next couple of weeks to get this work done, but I need to have in principle go-ahead before I start, because from the last time I took a stab at it, it got quite involved.

barakmich:
@kortschak That sounds great to me. That's also how `database/sql` does errors, so that makes a lot of sense.

I've been writing an alternative logging framework over at https://github.com/coreos/pkg/tree/master/capnslog with some of what we'd like for etcd et al. Returning errors is still the right thing where possible, but I took your "configurable logging" statement and played with the idea a bit, as well as having log entry objects which may contain more useful/important data which could be formatted.

SeanDolphin:
Having an Err() function on iterators in particular is a perfectly fine solution and is actually suggested to do so by Rob in http://blog.golang.org/errors-are-values

andrew-d:
So, I've sent a PR to this branch on @barakmich's repository with some initial docs.  For what my 2¢ are worth, I think that adding an `Err()` method on iterators is a good idea, and I'd be happy to help out with that too - @kortschak, would you mind if I threw something together to at least use as a starting point?  I'll admit that I'm not super-familiar with the cayley codebase - I've only found it recently - but I can probably get an initial attempt together.

kortschak:
Those design principles look like what I was thinking (I had the name "clog" in my mind).

On 15/04/2015, at 6:36 AM, "Barak Michener" <notifications@github.com<mailto:notifications@github.com>> wrote:

I've been writing an alternative logging framework over at https://github.com/coreos/pkg/tree/master/capnslog with some of what we'd like for etcd et al. Returning errors is still the right thing where possible, but I took your "configurable logging" statement and played with the idea a bit, as well as having log entry objects which may contain more useful/important data which could be formatted.

kortschak:
Now might be a good time to drop the ResultTree method.

andrew-d:
Okay, so, other than documentation, is there anything else that needs to be done before this PR can get merged?  I'll freely admit that I'm not super-great at docs, but if that's the only remaining thing blocking this PR, I should be able to find some time this week to write some.

kortschak:
I think ResultTree should go away before the API is published, but apart from that, I'm happy for this to be merged. I have some further changes to error handling to make, but those can happen after this (thanks for making that so much easier).

barakmich:
I've been rereading this and went over it again this weekend, and I think the reason this PR has been a slog is because something seemed off. So let's set a list of things to do before merging:

- [ ] Refactor such that root repo gets the API and `/cmd/cayley` is the binary
  - I've been thinking about it, and it's the right ultimate approach.
  - [ ] Move the integration test into `/integration` or similar
- [ ] Create a `Cursor` interface in `/graph`, of which Path is an implementation, with an appropriate method that takes the place of ApplyMorphism in the PR as written.
  - This is in pursuit of both #121 (and #232, as I've been thinking about it)
  - This can be expanded later
- [ ] Remove ResultTree

Optional:
  - [ ] Test/Build scripts

andrew-d:
Okay, so, `ResultTree` has been removed.  I can submit an additional PR that refactors the layout of things, but before I do that: do you want a particular API as the "top-level" package, or are you okay with leaving the API spread out through the packages it's in now (`graph`, `quad`, etc.)?

barakmich:
Most of the API should stay where it is. Extending the Go API (to simplify number of imports and so on) can now happen in the root. This may constitute subpackages, helper functions, etc... the goal being that my example above (which currently includes 5 imports) will eventually make it down to just `github.com/google/cayley`

So, here's my deal: make a PR to move the current root to `/cmd/cayley` (ie, point 1) and I'll do point 2 (hooks to move the handles for this up a few levels)

kortschak:
SGTM. Thanks people.

barakmich:
Alright, I'm happy to merge this. Still needs some help, and puts more pressure on #121 -- but we can distribute that work if this PR isn't blocking.

Incidentally, it's looking a lot better now. `/cmd/cayley` was the right move.

barakmich:
My little test code now looks like this:

```go
package main                                                                                                                                                                                   

import (
  "log"

  "github.com/google/cayley"
)

func main() {
  store, err := cayley.NewMemoryGraph()
  if err != nil {
    log.Fatalln(err)
  }
  store.AddQuad(cayley.Quad("food", "is", "good", ""))
  p := cayley.StartPath(store, "food").Out("is")
  it := p.BuildIterator()
  for cayley.RawNext(it) {
    log.Println(store.NameOf(it.Result()))
  }
}
```

kortschak:
Very nice. LGTM

Is it reasonable to use this API to build the js API on - I haven't look at that for a while? If we can, then we reduce duplication and ensure we have both APIs working the same way.

barakmich:
Absolutely. Deduplication would be a great cleanup. Merging to unblock.

imports.go
```diff
@@ -0,0 +1,46 @@
...
+	return quad.Quad{subject, predicate, object, label}
+}
+
+func NewMemoryGraph() (*Handle, error) {
+	qs, err := graph.NewQuadStore("memstore", "", nil)
+	if err != nil {
+		return nil, err
+	}
+	qw, err := graph.NewQuadWriter("single", qs, nil)
+	if err != nil {
```
neonstalwart: i'm curious... if you get into this conditional, should you `qs.Close()`?

kortschak: Close on a memstore.Quadstore is a no-op.

neonstalwart: ok, but *if* it did something then i assume you would need to close it, right?

e.g. does  https://github.com/google/cayley/blob/680bb244d2c7e28dcbf9abe3229f2954dc030130/db/db.go#L45 need to `qs.Close()`?

kortschak: I would say so. The close may not be effective depending on the error, but it should be tried. I really must find some time to clean the error handling so we can deal with these things properly.

neonstalwart: thanks, i was just wanting to understand. i appreciate the response.

PR #210, opened 2015-02-09
Make query/... interfaces more idiomatic

kortschak:
Also revert the data type returned by queries to interface{} (the change made sense at the time).


barakmich:
session.go is so much cleaner. And the query list (now that it's sorted and checked) while the major line contributor is obviously the right next step from a previous change. LGTM

Issue #200, opened 2015-01-21
Case insensitive queries


ecesena:
I'm playing with Cayley and Freebase and one of the biggest "ux" issues is the lack of a case insensitive queries in Cayley. 

I was wondering if this is a feature you could see added to Cayley (maybe via a case insensitive iterator?), or if there is any specific design/implementation choice that goes against it. 

(I've read the guidelines and TODO, I didn't find anything related, sorry if I missed something.)

robertmeta:
Great feature request -- moved to our feature request forum: https://discourse.cayley.io/t/case-insensitive-queries/231

PR #212, opened 2015-02-12
Add support for string collation

LAlbertalli:
This is just an initial pull requests to implement the support for string collation (somewhat similar to what exists in SQL Databases) that address the issue #200. It is just a starting point, I have some doubts that should be addressed.
The collation is supported only on BoltDB for now but the implementation on other storage engine should be easy.

## Basic idea
Right now the indexes to search in the database use the hash of the string. In order to support the string collation, before hashing the stream, the string is converted to its comparison key using the used collator. If no collation is specified the storage engine works as before.

## Configuration
In order to configure the collation the following options have been added to the configuration file:
#### **`collation_type`**

  * Type: String
  * Default: ""

  The collation to be used for string comparison by the database. The string should be a valid BCP 47 language tag. If empty no collation is used.

#### **`collation_options`**

  * Type: List of String
  * Default: []

  A list of option tag to be applied to the selected collation. Valid options are: "IgnoreCase", "IgnoreDiacritics", "IgnoreWidth", "Loose", "Force", "Numeric"

The same options are available on the command line with `-collation_type` and `-collation_options`. The later is a comma separated string.

## Open questions

* Right now every string is collated. A proper implementation should implement a different behavior based on the type of the object as defined by RDF standard (http://www.w3.org/TR/n-quads/#n-quads-language):
  * IRI should not be collated
  * string literals should be collated according to the specified language or the default if it is not defined.
  * string that represents a different datatype should not be collated

  I'm not happy with this level of complication because it would complicate the query. A differentiation between IRIs (that should NOT be collated) and string (to be collated) could be interesting even if I think that right now the parser drops the the angular bracket so more work is required.
* If the db is opened with a different collation there's no error but the behavior could be unexpected. Probably the collation should be stored in the db and checked open start.
* The Collator is now implemented in the graph package but should be probably moved in a different package.
* Only one collation is supported right now. Multiple collation support (e.g: specifying it at connection level) could be used but it should be properly designed and I don't see a use case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,8 @@ install:
   - go get github.com/boltdb/bolt
   - go get gopkg.in/mgo.v2
   - go get gopkg.in/mgo.v2/bson
-  - go get golang.org/x/text
+  - go get golang.org/x/text/collate
+  - go get golang.org/x/text/language
 
 script: go test -v ./...
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - go get github.com/boltdb/bolt
   - go get gopkg.in/mgo.v2
   - go get gopkg.in/mgo.v2/bson
+  - go get golang.org/x/text
 
 script: go test -v ./...
 
--- a/cayley.go
+++ b/cayley.go
@@ -30,6 +30,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"time"
+	"strings"
 
 	"github.com/barakmich/glog"
 
@@ -65,6 +66,8 @@ var (
 	port               = flag.String("port", "64210", "Port to listen on.")
 	readOnly           = flag.Bool("read_only", false, "Disable writing via HTTP.")
 	timeout            = flag.Duration("timeout", 30*time.Second, "Elapsed time until an individual query times out.")
+	collationType      = flag.String("collation_type", "", "Specify the collation to use")
+	collationOptions   = flag.String("collation_options", "", "Comma separated list of collation options")
 )
 
 // Filled in by `go build ldflags="-X main.Version `ver`"`.
@@ -141,6 +144,14 @@ func configFrom(file string) *config.Config {
 	}
 
 	cfg.ReadOnly = cfg.ReadOnly || *readOnly
+
+	if cfg.CollationType == "" {
+		cfg.CollationType = *collationType
+	}
+
+	if len(cfg.CollationOptions) == 0 && *collationOptions != "" {
+		cfg.CollationOptions = strings.Split(*collationOptions,",")
+	}
 
 	return cfg
 }
--- a/cayley.go
+++ b/cayley.go
@@ -187,6 +187,7 @@ func main() {
 		handle *graph.Handle
 		err    error
 	)
+	graph.InitCollator(cfg)
 	switch cmd {
 	case "version":
 		if Version != "" {
--- a/cayley.go
+++ b/cayley.go
@@ -187,7 +187,11 @@ func main() {
 		handle *graph.Handle
 		err    error
 	)
-	graph.InitCollator(cfg)
+	err =graph.InitCollator(cfg)
+	if err != nil{
+		glog.Errorln(err)
+		os.Exit(1)
+	}
 	switch cmd {
 	case "version":
 		if Version != "" {
--- a/config/config.go
+++ b/config/config.go
@@ -34,6 +34,8 @@ type Config struct {
 	ReadOnly           bool
 	Timeout            time.Duration
 	LoadSize           int
+	CollationType      string
+	CollationOptions   []string
 }
 
 type config struct {
@@ -47,6 +49,8 @@ type config struct {
 	ReadOnly           bool                   `json:"read_only"`
 	Timeout            duration               `json:"timeout"`
 	LoadSize           int                    `json:"load_size"`
+	CollationType      string                 `json:"collation_type"`
+	CollationOptions   []string               `json:"collation_options"`
 }
 
 func (c *Config) UnmarshalJSON(data []byte) error {
@@ -66,6 +70,8 @@ func (c *Config) UnmarshalJSON(data []byte) error {
 		ReadOnly:           t.ReadOnly,
 		Timeout:            time.Duration(t.Timeout),
 		LoadSize:           t.LoadSize,
+		CollationType:      t.CollationType,
+		CollationOptions:   t.CollationOptions,
 	}
 	return nil
 }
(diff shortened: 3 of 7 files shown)


LAlbertalli:
It seems the Travis-CI environment doesn't support the packages in golang.org/x path. I'm investigating

LAlbertalli:
Now the Travis-CI works, was a mistake on my side.

barakmich:
@LAlbertalli It took me a bit to pick this up and see where you're going with it.

I think I get what your intent is; however if I'm not mistaken, this makes the entire database follow the collation rules (ie, case insensitivity, et al?)

Now nodes "A" and "a" are the same. See also the other collation options. A person with lastName "Mueller" and one with lastName "Müeller" are not the same. Though they may want to be searched/traversed that way, they are not, canonically, the same, and representing them in the database as the same node is going down a troubling path.

I completely understand why #200 exists and it's a useful ask. In fact, I think it's part of a more general problem/pattern: the textual index. 

So yeah, we'd need a new index. Which can leverage the backing store where possible. Where not possible, we'd need to (a) provide one (say in a new Bolt bucket) or (b) claim that this feature is unsupported for some particular backends (which I don't like, but I can't conclusively prove is unavoidable). Many backends would do this for us -- an SQL backend could leverage LIKE for example. For memory/Bolt/LevelDB -- we'd either need to invent one per (using a similar trick to what you've written here, for canonicalization) or use a text-index library (My sketchbook contains a mention to look at [bleve](https://github.com/blevesearch/bleve) )

This would also involve a value-iterator of some form -- all values that match (in the appropriate ways) the value at hand.

On the query level, this would hopefully manifest as
```
g.V("Mueller")  // (set of exactly one, "Mueller")
g.Find("mueller") // (set of potentially many, case insensitive and whatnot)
```

I'm sketching it out in this comment, but I also really appreciate your documentation style. If this sounds cool to you (and I'm really happy to help here with some subfeatures) perhaps we start a design issue?

LAlbertalli:
@barakmich yes, issue #200 was opened from a colleague of mine, so I know exactly the issue. And after seeing the Cayley architecture I thought that a full collation support was better than just case insensitive search.
I understand where you came from in your proposal and I like it but let me go a little bit to the core of the problem before discussing the solution. "Straße" and "Strasse" are two different word for an English speaking person but are the same word for a German one. And "Obama" seems a single node but could reference the President of the United States or the First Lady. And maybe "A" and "a" are the same node but "Apple" and "apple" are not (one is the company, the other the fruit). I'm making real examples on which my colleagues are working right now. If we use Cayley for storing an entities graph for text analysis the problem of different word that collates to the same one is just a subset of the disambiguation problem. Our experience on text analysis with regard to the collation (we use Python that doesn't support collation, and MySQL DB which it does) is that the better approach is having your instrument collations aware and manage the ambiguity that arises at the application level were you could apply heuristic specific to your field. Plus it is easier to run collated comparison and then filter the result to be bitwise than the other way around (the possible string to test or search are just huge).
Obviously, if the collations doesn't apply to the data stored (I don't know, a DNA sequence?) the database could be configured without collation.
Adding an index for text search is a compelling idea, I like it and was the first solution I analyzed. But I have discarded it for these reasons:
* Increase the code complexity (I assume the idea is to have an index that maps the single collated string to the multiple not collated one. From a certain point of view `g.Find("mueller) == g.V("Mueller","Müeller"...)` ),
* Slow down the find query having to do an additional lookup,
* Increase the space required for the database around 10-15% if I estimated it right,

I suspect that few people will mix V and Find in their usual workflow and the find could be easily transformed in a V (the original string is not transformed, only the indexes are collated) while it is not true for the opposite.

Also, my expectation for a Find syntax is something more similar to the SQL LIKE or a regexp than just a collated search. My guess is that a Find feature would be cool but should solve a different set of problems than just being sure that "Mueller" == "Müeller" for the search in the database (that in the end is the meaning of a collation).

But these are obviously my considerations, I'm more than happy to hear the suggestions of other Cayley users.

Bye
L.

barakmich:
I talked this PR over with some friends of mine from the Freebase world. (I do take it very seriously)

While I still think that a textual index would be cool (and worth the growth in size) there's another place you could hook into the Extract, Transform, and Load.... In the Transform, instead of the Load.

Right now we parse quads either strictly or non-strictly, but at the end of the day, these are functions from file to list of quads. 

You could get equivalent behavior to this PR, without intentional loss of data or strange indexing, and without an extra index. Essentially, you could imagine functions which take sets of quads, and transform them into other sets of quads. Think  `io.Reader` wrapped in `bufio`

Creating a Transform interface (name of your choice) which takes a list of quads and returns a list of quads -- before they get written -- would be kind of nice. In this case, the transform is simple; lowercase/canonicalize the strings (here you could also do appropriate things for RDF types, as you mention). More advanced versions of this could reconcile data, pull related data, deduplicate data... or any number of things.

You could also chain these, which simplifies some of the fanout of options you have.

Indexing progresses as normal.

What do you think?

kortschak:
This makes very much more sense to me. The idea of doing things magical inside the engine was making me a little queazy. I'd be very much happier if transforms happened at this boundary instead.

barakmich:
@kortschak Incidentally, this may ultimately provide a way out of the nquads/cquads weirdness -- as cquads are a relatively simple transform on proper RDF NQuads.

kortschak:
That's true, but nquads are faster to parse because the grammar is more
restrictive. The state graph of cquads is insane.

LAlbertalli:
If I understand the comment, the idea is to define an interface to specify how to transform the data during the loading. This idea is for sure a good one and allow many future extension. But regarding the specific problem I'm trying to address, I see some limitations:
* the same transformations should be applied to the data used for querying. I see two possibility:
  * moving the transformation on the business logic that runs the query, with the side effect of complicating the query process.
  * having the same transformation applied also to the query, but this will complicate the interface.
* Loss of information. If for instance I introduce a canonicalize and a lower case transformation, in the node bucket I will find for instance "new york" instead of "New York" and "Strasse" instead of "Straße" and that would be the data returned to me. Depending on the context this loss of information could be acceptable or not.

I understand the issue over loosing informations in the index but I feel that losing it, also if intentionally, in the overall data is an even bigger concerns (let me know if I have misunderstood the proposal, obviously). I think I should go back to the drawing board to try to address all the concerns. I don't have much time right now but I try to came up with something by the end of the week.

barakmich:
Hey! Did you come up with an alternative?

LAlbertalli:
Hi Barak,

Sorry, my fault. In this moment other priorities has emerged so I hadn't had time to think at this issue.
I think the best move right now is to to close the Pull Requests and I will maintain my separate branch with the collator for my use case.

Bye
L.

barakmich:
@LAlbertalli You can do that, because that's open source, but I don't want to discourage contributions :cry:  
If you have other PRs, you can send them, of course. And hopefully with text indexing (or similar) we can reevaluate in the future.

LAlbertalli:
I know and I'm willing to contribute more features. Just in this moment I have to refocus on new features and Cayley with this patch is enough for now, so I can't devote my time to this feature.

I will probably come back on this in few months...

Thanks
L.

PR #214, opened 2015-02-18
Added Emit Alias 

sagivo:
Emit() = E()


barakmich:
I'm declining this, for a simple reason: `E` is overloaded with `Edge`, and this change doesn't buy us much in terms of usability, as Emit is not nearly as common as some other operations.

However, you have the right idea of embedding JS. Write a cool JS function attached to `graph`, and I'll take a look.

PR #217, opened 2015-02-21
Rewrite keys into concrete types, remove key package

### LINKED ISSUE #215: Concretize PrimaryKey, bring it into graph
/cc @panamafrancis 

I've been writing single master replication and the current PrimaryKey interface is both (a) subtly broken (one line fix. See if you can find it) and (b) ridiculously hard to use. Most times you end up treating it as your preferred type of identifier.

On the one hand, I'm curious what UUIDs buy you over monotonic ids, but even assuming that, there's a better way.

If this were C, it'd be a union type. As this is not C, a type with a int8 type field and a few sub fields, per usable key implementation. 

Why not use an interface? Because it can't be unmarshaled. And putting Int() everywhere it actually should be used, straight up, is a major pain.

So I'll be making this change shortly. UUIDs will still have a hook, so nothing gets lost, but a lot is gained. 

Bringing it into graph (as it will now be a one-stop type) is just simplifying imports.

barakmich:
Fixes #215


kortschak:
Out of curiosity, what is the reason for bringing this into graph? Purely to reduce imports?

barakmich:
Yeah, just to reduce imports. I don't think there will be very many alternative keys and since the interface already lived in graph it was a pretty natural move.

Importantly, it's now possible to Marshal/Unmarshal `graph.Delta`s (which contain `PrimaryKey`s) . Which is quite useful.

kortschak:
Fair enough. LGTM

PR #219, opened 2015-02-23
Primary Keys: Implemented unique keys 

panamafrancis:
Also did the (Un)Marshal interface implementations. All tested.

In regards to the MarshalJSON() & UnmarshalJSON() implementations, I looked over a couple of alternatives such as using a map[string]interface{} or creating a separate struct just for JSON marshalling but i decided just exporting the members (KeyType, SequentialID...) made more sense as the other approaches had high maintenance costs.

I would rather do this commit with the rest of my changes but i think it's now more appropriate to do separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - go get github.com/boltdb/bolt
   - go get gopkg.in/mgo.v2
   - go get gopkg.in/mgo.v2/bson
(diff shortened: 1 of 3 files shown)


panamafrancis:
Or.. would it be preferred if i split it into two commits?

panamafrancis:
I'm going to close this PR and commit changes with the GAE PR, makes more sense for now.

PR #226, opened 2015-04-05
graph: Cassandra backend

barakmich:
#### Adds a Cassandra backend.

This has been a WIP for awhile, but since it seems more or less stable to my testing (and the integration test as well) I think it's about time to make it a PR. 

Here's the thing: Cassandra is a nicer production backend than Mongo IMO.

Why did I write a Mongo backend to begin with? To prove that, by offloading the work to a sharded backend, you could have a graph layer that worked. Mongo, for all its faults, has a very easy ramp and is an easy way to get started and show off a distributed backend.
#### Things that need to be done before submission:
- Docs for the Cassandra `db_options` (notably, consistency levels)
#### Followup PRs:
- Cassandra Writer -- following the DB options provided, use the metadata table to synchronize multiple writers using Cassandra's underlying Paxos implementation (UPDATE-IF in CQL). This means (ideally) multiple Cayley writers and therefore less state/coordination.
#### Stats:

Quick localhost benchmarking:

```
$ go test --bench=. -v . --backend=cassandra --exists
=== RUN TestQueries
--- PASS: TestQueries (8.03s)
=== RUN TestDeletedAndRecreatedQueries
--- PASS: TestDeletedAndRecreatedQueries (7.93s)
=== RUN TestDecompressor
--- PASS: TestDecompressor (0.00s)
PASS
BenchmarkNamePredicate       500       3481455 ns/op
BenchmarkLargeSetsNoIntersection           3     353195996 ns/op
BenchmarkVeryLargeSetsSmallIntersection        2     899606857 ns/op
BenchmarkHelplessContainsChecker           1    1364770289 ns/op
BenchmarkHelplessNotContainsFilms          1    2403209044 ns/op
BenchmarkHelplessNotContainsActors         1    1856315458 ns/op
BenchmarkNetAndSpeed          30      39402451 ns/op
BenchmarkKeanuAndNet         100      21139925 ns/op
BenchmarkKeanuAndSpeed        50      30976309 ns/op
BenchmarkKeanuOther       10     198898765 ns/op
BenchmarkKeanuBullockOther         5     347030111 ns/op
ok      github.com/google/cayley    83.672s
```

Not too shabby, all things considered.


proteneer:
+1 this is awesome!

barakmich:
@proteneer Thanks! I see Github has you out in NYC -- if you'd like to chat ever over coffee, I'm always looking for contributors and am happy to plan offline.

graph/cassandra/create_quadstore.go
```diff
@@ -0,0 +1,137 @@
...
+	CREATE TABLE metadata (
+		meta_key text PRIMARY KEY,
+		value bigint,
+	)
+	`).Exec()
+	if err != nil {
+		glog.Errorln("Could not create table metadata:", err)
+		return err
+	}
+	session.Close()
```
jzelinskie: why not defer?

graph/cassandra/quadstore.go
```diff
@@ -0,0 +1,378 @@
...
+	keyspace := DefaultKeyspace
+	if val, ok := options.StringKey("keyspace"); ok {
+		keyspace = val
+	}
+	consistencyStr := DefaultReadConsistency
+	if val, ok := options.StringKey("consistency"); ok {
+		consistencyStr = val
+	}
+	cluster.Consistency = getConsistency(consistencyStr)
+	cluster.Keyspace = keyspace
```
jiminoc: a few more handy options for C* performance for gocql is controlling number of conns and streams and the discover hosts option 
cluster.NumConns = 2
cluster.NumStreams = 16
cluster..DiscoverHosts = true

discover hosts true should be safe, conns and streams can be optional params passed in

graph/cassandra/quadstore.go
```diff
@@ -0,0 +1,378 @@
...
+	batch.Cons = qs.writeConsistency
+	counterBatch := qs.sess.NewBatch(gocql.CounterBatch)
+	counterBatch.Cons = qs.writeConsistency
+	newSize := qs.size
+	newHorizon := qs.horizon
+	for _, d := range deltas {
+		if d.Action == graph.Add {
+			newSize++
+			qs.addDeltaToLog(d, batch, newSize)
+			qs.addQuadToBatch(d, batch, counterBatch)
```
jiminoc: The batching here is somewhat dangerous from a performance perspective. The issue is that you're batching multiple tables/primary keys which means that the coordinator that gets the request has to go to multiple nodes to satisfy the node that holds each key. That puts a lot of pressure on the coordinator node that has to accept that request. It could end of actually having to hit all the nodes in your cluster (depending on batch and cluster size) and since it's Logged that log has to replicate n nodes as well. So it's a very expensive write in a moderate performance use case. 

A better approach with batching is to group only the writes that hit a single vertex(primary key) together if possible and make a batch out of that and single exec the others which will allow gocql to route the primary key to the appropriate node in the cluster if DiscoverHosts is set. It could be that  your batches are 2-5 writes instead of a whole block of random keys. 

I'm also assuming you're using batching for consistency across tables and not for performance reasons?

graph/cassandra/quadstore.go
```diff
@@ -0,0 +1,378 @@
...
+	batch := qs.sess.NewBatch(gocql.LoggedBatch)
+	batch.Cons = qs.writeConsistency
+	counterBatch := qs.sess.NewBatch(gocql.CounterBatch)
+	counterBatch.Cons = qs.writeConsistency
+	newSize := qs.size
+	newHorizon := qs.horizon
+	for _, d := range deltas {
+		if d.Action == graph.Add {
+			newSize++
+			qs.addDeltaToLog(d, batch, newSize)
```
jiminoc: I'm assuming this log is for debugging purposes?
Would be nice to make that configurable on or off, off by default since you're adding an extra write for each operation as well as the storage size of that data for large graphs. Also don't see a TTL on the logging statement so it appears the logs stick around indefinitely with unbounded growth?

graph/cassandra/create_quadstore.go
```diff
@@ -0,0 +1,137 @@
...
+		return err
+	}
+	cluster.Consistency = gocql.All
+	session, err := cluster.CreateSession()
+	if err != nil {
+		glog.Errorln("Could not create a Cassandra graph:", err)
+		return err
+	}
+	err = session.Query(`
+	CREATE TABLE log (
```
jiminoc: this will default to size tiered compaction which would be bad for ttl'ing out data. It might be nice to have this as DateTieredCompactionStrategy with a reasonable max_size so that after n days if all the keys are expired C* just has to rm the SSTable vs having to do a compaction operation on it. For log type data that's a natural fit for a time-series type model.

mikeatm:
Im unfamiliar with column stores like cassandra,  so im going to ask, how specific is this integration, has anyone  any thoughts on  whether it will work against the likes of  https://github.com/scylladb/scylla  that claim 100% compatibility with cassandra?

ccl0326:
What's the state of this PR now?

graph/cassandra/create_quadstore.go
```diff
@@ -0,0 +1,137 @@
...
+		action text,
+		subject text,
+		predicate text,
+		object text,
+		label text,
+	)
+	`).Exec()
+	if err != nil {
+		glog.Errorln("Could not create table log:", err)
+		return err
```
SamuelRamond: you need to `session.Close()` or `defer session.Close()` L:35

colek42:
What needs to get done with this?  I would like to get this to a state where it can be merged and can donate some cycles if there is a clear path.

dennwc:
@colek42 It needs to be rebased onto master at least. Also, if I remember correctly, @barakmich wanted to rewrite it to some degree.

barakmich:
Closing my own for age. Requires a rewrite.

Issue #157, opened 2014-09-20
improve error handling and propagation


kortschak:
At the moment cayley does not return errors in many failable cases. This has arisen because new back ends have been added that have failable cases that did not exist in the original stores. Also it is partly historical.

This set of changes may require alterations to the Iterator and QuadStore interfaces.

iddan:
@dennwc is this still relevant?

dennwc:
It should be solved by a recent iterator refactoring. Closing.

PR #229, opened 2015-04-15
Add Err() and Close() methods to Iterators

andrew-d:
This PR does the following:
- As mentioned in #208, add an `Err()` method to all Iterators.
- Adds tests that the various iterators pass through the underlying Err value.
- Makes the `Close()` method on Iterators return an error.
- Remove a bunch of uses of `glog.Fatal*` calls, instead either propagating the errors, or calling `panic` (when it's a programming bug).
- Try to handle errors in other locations that use `Next()`.

cc @kortschak 

Updates #157

### DIFF
--- a/graph/bolt/iterator.go
+++ b/graph/bolt/iterator.go
@@ -106,10 +106,11 @@ func (it *Iterator) Clone() graph.Iterator {
 	return out
 }
 
-func (it *Iterator) Close() {
+func (it *Iterator) Close() error {
 	it.result = nil
 	it.buffer = nil
 	it.done = true
+	return nil
 }
 
 func (it *Iterator) isLiveValue(val []byte) bool {
--- a/graph/bolt/iterator.go
+++ b/graph/bolt/iterator.go
@@ -46,6 +46,7 @@ type Iterator struct {
 	dir     quad.Direction
 	qs      *QuadStore
 	result  *Token
+	err     error
 	buffer  [][]byte
 	offset  int
 	done    bool
@@ -170,6 +171,7 @@ func (it *Iterator) Next() bool {
 		if err != nil {
 			if err != errNotExist {
 				glog.Errorf("Error nexting in database: %v", err)
+				it.err = err
 			}
 			it.done = true
 			return false
@@ -182,6 +184,10 @@ func (it *Iterator) Next() bool {
 		return false
 	}
 	return true
+}
+
+func (it *Iterator) Err() error {
+	return it.err
 }
 
 func (it *Iterator) ResultTree() *graph.ResultTree {
--- a/graph/bolt/iterator.go
+++ b/graph/bolt/iterator.go
@@ -316,3 +316,5 @@ func (it *Iterator) Stats() graph.IteratorStats {
 		Size:         s,
 	}
 }
+
+var _ graph.Nexter = &Iterator{}
--- a/graph/bolt/iterator.go
+++ b/graph/bolt/iterator.go
@@ -46,11 +46,11 @@ type Iterator struct {
 	dir     quad.Direction
 	qs      *QuadStore
 	result  *Token
-	err     error
 	buffer  [][]byte
 	offset  int
 	done    bool
 	size    int64
+	err     error
 }
 
 func NewIterator(bucket []byte, d quad.Direction, value graph.Value, qs *QuadStore) *Iterator {
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -136,7 +136,7 @@ type Iterator interface {
 	Describe() Description
 
 	// Close the iterator and do internal cleanup.
-	Close()
+	Close() error
 
 	// UID returns the unique identifier of the iterator.
 	UID() uint64
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -100,7 +100,7 @@ type Iterator interface {
 	// Contains returns whether the value is within the set held by the iterator.
 	Contains(Value) bool
 
-	// Err returns the error (if any) encountered during iteration.
+	// Err returns any error that was encountered by the Iterator.
 	Err() error
 
 	// Start iteration from the beginning
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -177,6 +177,17 @@ func Next(it Iterator) bool {
 	return false
 }
 
+// Err is a convenience function that conditionally calls the Err method
+// of an Iterator if it is a Nexter.  If the Iterator is not a Nexter, Err
+// returns nil.
+func Err(it Iterator) error {
+	if n, ok := it.(Nexter); ok {
+		return n.Err()
+	}
+	glog.Errorln("Calling Err on an un-nextable iterator")
+	return nil
+}
+
 // Height is a convienence function to measure the height of an iterator tree.
 func Height(it Iterator, until Type) int {
 	if it.Type() == until {
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -155,8 +155,13 @@ type Description struct {
 
 type Nexter interface {
 	// Next advances the iterator to the next value, which will then be available through
-	// the Result method. It returns false if no further advancement is possible.
+	// the Result method. It returns false if no further advancement is possible, or if an
+	// error was encountered during iteration.  Err should be consulted to distinguish
+	// between the two cases.
 	Next() bool
+
+	// Err returns the error (if any) encountered during iteration.
+	Err() error
 
 	Iterator
 }
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -177,17 +177,6 @@ func Next(it Iterator) bool {
 	return false
 }
 
-// Err is a convenience function that conditionally calls the Err method
-// of an Iterator if it is a Nexter.  If the Iterator is not a Nexter, Err
-// returns nil.
-func Err(it Iterator) error {
-	if n, ok := it.(Nexter); ok {
-		return n.Err()
-	}
-	glog.Errorln("Calling Err on an un-nextable iterator")
-	return nil
-}
-
 // Height is a convienence function to measure the height of an iterator tree.
 func Height(it Iterator, until Type) int {
 	if it.Type() == until {
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -100,6 +100,9 @@ type Iterator interface {
 	// Contains returns whether the value is within the set held by the iterator.
 	Contains(Value) bool
 
+	// Err returns the error (if any) encountered during iteration.
+	Err() error
+
 	// Start iteration from the beginning
 	Reset()
 
@@ -159,9 +162,6 @@ type Nexter interface {
 	// error was encountered during iteration.  Err should be consulted to distinguish
 	// between the two cases.
 	Next() bool
-
-	// Err returns the error (if any) encountered during iteration.
-	Err() error
 
 	Iterator
 }
--- a/graph/iterator/all_iterator.go
+++ b/graph/iterator/all_iterator.go
@@ -103,6 +103,11 @@ func (it *Int64) Next() bool {
 	return graph.NextLogOut(it, val, true)
 }
 
+func (it *Int64) Err() error {
+	// This iterator should never error.
+	return nil
+}
+
 // DEPRECATED
 func (it *Int64) ResultTree() *graph.ResultTree {
 	return graph.NewResultTree(it.Result())
--- a/graph/iterator/all_iterator.go
+++ b/graph/iterator/all_iterator.go
@@ -55,7 +55,9 @@ func (it *Int64) Reset() {
 	it.at = it.min
 }
 
-func (it *Int64) Close() {}
+func (it *Int64) Close() error {
+	return nil
+}
 
 func (it *Int64) Clone() graph.Iterator {
 	out := NewInt64(it.min, it.max)
--- a/graph/iterator/all_iterator.go
+++ b/graph/iterator/all_iterator.go
@@ -160,3 +160,5 @@ func (it *Int64) Stats() graph.IteratorStats {
 		Contains:     it.runstats.Contains,
 	}
 }
+
+var _ graph.Nexter = &Int64{}
--- a/graph/iterator/all_iterator.go
+++ b/graph/iterator/all_iterator.go
@@ -106,7 +106,6 @@ func (it *Int64) Next() bool {
 }
 
 func (it *Int64) Err() error {
-	// This iterator should never error.
 	return nil
 }
 
--- a/graph/iterator/and_iterator.go
+++ b/graph/iterator/and_iterator.go
@@ -293,9 +293,9 @@ func (it *And) Close() error {
 
 	err := it.primaryIt.Close()
 	for _, sub := range it.internalIterators {
-		serr := sub.Close()
-		if serr != nil && err == nil {
-			err = serr
+		_err := sub.Close()
+		if _err != nil && err == nil {
+			err = _err
 		}
 	}
 
--- a/graph/iterator/and_iterator.go
+++ b/graph/iterator/and_iterator.go
@@ -29,6 +29,7 @@ type And struct {
 	primaryIt         graph.Iterator
 	checkList         []graph.Iterator
 	result            graph.Value
+	err               error
 	runstats          graph.IteratorStats
 }
 
@@ -153,7 +154,14 @@ func (it *And) Next() bool {
 			return graph.NextLogOut(it, curr, true)
 		}
 	}
+	if err := graph.Err(it.primaryIt); err != nil {
+		it.err = err
+	}
 	return graph.NextLogOut(it, nil, false)
+}
+
+func (it *And) Err() error {
+	return it.err
 }
 
 func (it *And) Result() graph.Value {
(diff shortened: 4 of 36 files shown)


kortschak:
I had planned on adding Err() to Iterator, but this looks OK too - I can't really distinguish which is better. Can you provide pros for this approach? (On the face of it, this looks a bit neater in how it will pan out, but I'd like to hear your motivation here). Thanks for picking this up BTW.

Will look in more detail later.

andrew-d:
Well, the original reason that I added `Err()` to `Nexter` was mostly because I assumed the only reason you'd call it is when iterating using `Next()`.  However, if there are other things that might need it, I'd assume that I can just move it over to the `Iterator` interface.  This PR is still rather work-in-progress :smile:

kortschak:
Yeah. I think that's a plan. We can move it without concern if needed later. This does localise the feature to where it's needed AFAICS at the moment - when we dive into errors we may need expanding.

The Close method on graph.Iterator should return an error I think. Would you add that please.

andrew-d:
Okay, so, I've got most of the major iterators fixed, one per commit.  I only have two more to finish, and I'd appreciate some input:

**Materialize**  
Essentially, this runs the entire subiterator and caches the result.  Should this return an error if the subiterator returns one (i.e. return no results), or should it iterate over the results it has cached and then return the error?

**LinksTo**  
I'm not sure I understand what this iterator does, so I'm a bit leery of just randomly throwing error handling in.

And finally, a more general question - do you think it's worth having the various iterators check if the current error is non-nil in `Next()` and, if so, returning false?  The alternative being, of course, telling the user that "you must check for non-nil Err if Next returns false".

And finally, sure, I can add the error return to `Close()`.

kortschak:
My view is that an Err() != nil state means the Next call failed, and the user should find out why (this is provoked by not getting anything back). This would be consistent with Materialize returning no results on failure.

The justification for this is that partial results may not be safe, and we shouldn't be fostering unsafe behaviour.

Wait for @barakmich to give his view though.

andrew-d:
@kortschak Okay, added the `Close()` method.  One thing that is worth noting is that, for iterators that aggregate from underlying iterators (e.g. `Or`), if an error occurs, only the first one is returned.  I did my best to document this, but thoughts are appreciated.

kortschak:
I think it's reasonable to only return the first. We can extend that at a later date if it's deemed necessary - I guess it really depends on the needs of the backend (the authors of those may want to add handling for multiple error returns via a []error type that implements error).

barakmich:
So I'd say that it's probably wise to put Err() on the iterator interface. A good example of this is HasA.Check() which can Next() its way through a subiterator and possibly have an error therein. The good news is it's a slight interface change away. :)

I'm +1 with @kortschak that partial results may not be safe, and consistent with Materialize returning no results on failure. Not to conflate it with no results on success. ;) 

The rule of thumb being, as an iterator, if a subiterator has an error, we should catch it and error ourselves. Whatever that first error happens to be (extending to []error if necessary, as discussed). Materialize is actually simple, because it's exactly this -- the only time it produces errors of its own should be if memory allocation fails or something equivalently implementation-detail-oriented, because it's the same set of things as the subiterator, just pre-fetched.

LinksTo is, to explain a bit of what it does, the set of all links that are "links to" the subiterator of nodes. HasA is the set of nodes that a subiterator of links has a specific field of. (`foo bar baz .` "has a" subject being `foo`)

Also, thank you so much for digging in with both hands. This is awesome and cleans up a lot of the code :+1:

andrew-d:
@barakmich @kortschak Okay, I've added all the various Err / Close methods.  It does appear that I've broken the tests somewhat - `go test -v ./query/gremlin` is a simple example of what no longer works.  I've tried to avoid any functional changes, but clearly I missed something :smile: 

I'll go through the code again sometime today, but if you have any suggestions, it would be appreciated.

andrew-d:
Aha!  Okay, that last commit fixed it.  `git bisect` is great :smiley: 

So: all tests pass, all `Iterator`s have an `Err()` method, and all `Close()` calls return an error.  @barakmich, the last question I have for you relates to error handling in other parts of the code.  Currently, I more or less assumed that errors really only occur during iteration (i.e. during `Next()`).  Is there any other parts of the code you want me to fiddle with?

kortschak:
Have a look for any glog.Fatal calls. Function that contain them, should return an error if they don't already.

barakmich:
And take a look for errors related to NextResult() and Check() -- those are iteration-related.

andrew-d:
@kortschak Will do.

@barakmich So, maybe I'm missing something, but I don't see any functions called either `NextResult` or `Check` in the code?

andrew-d:
@kortschak, @barakmich: Okay, I removed a bunch of `glog.Fatal*` calls.  I converted a couple to `panic`s - stuff that looked like a programming bug, not necessarily an error.

Also: I added more error handling in the HasA iterator.  I think that, as of now, all uses of `Next()` in the code are either:
- In `Next()` methods themselves
- In tests
- Or properly handle errors.

Feel free to double-check - here's what I've been using: `ag '\.Next\(' -G '^(?!(.*_test\.go)$)'`

kortschak:
NextResult and Check got renamed a while ago to NextPath and (from dim memory) Contains.

barakmich:
Correct. Old dog, new tricks. ;)

andrew-d:
@barakmich Done!  That being said, I noticed [something strange](https://github.com/google/cayley/pull/229/files#diff-aff85d12107d6522135107af61084319R199) - any idea if that result is supposed to be used?

graph/iterator.go
```diff
@@ -100,6 +100,9 @@ type Iterator interface {
 	// Contains returns whether the value is within the set held by the iterator.
 	Contains(Value) bool
 
+	// Err returns the error (if any) encountered during iteration.
```
kortschak: Err returns any error that was encountered by the Iterator.

graph/iterator/all_iterator.go
```diff
@@ -103,6 +105,11 @@ func (it *Int64) Next() bool {
 	return graph.NextLogOut(it, val, true)
 }
 
+func (it *Int64) Err() error {
+	// This iterator should never error.
```
kortschak: I don't think this comment is necessary.

graph/iterator/and_iterator.go
```diff
@@ -255,13 +276,25 @@ func (it *And) cleanUp() {}
 // Close this iterator, and, by extension, close the subiterators.
 // Close should be idempotent, and it follows that if it's subiterators
 // follow this contract, the And follows the contract.
-func (it *And) Close() {
+//
+// Note: as this potentially involves closing multiple subiterators, only
```
kortschak: Don't add this as a note, just say "It closes all sub iterators it can, but returns the first error it encounters." (cribbed from os.RemoveAll).

Also below where this is used.

graph/iterator/fixed_iterator.go
```diff
@@ -143,6 +145,11 @@ func (it *Fixed) Next() bool {
 	return graph.NextLogOut(it, out, true)
 }
 
+func (it *Fixed) Err() error {
+	// This iterator should never error.
```
kortschak: Drop comment.

graph/iterator/hasa_iterator.go
```diff
@@ -239,11 +262,20 @@ func (it *HasA) Stats() graph.IteratorStats {
 }
 
 // Close the subiterator, the result iterator (if any) and the HasA.
-func (it *HasA) Close() {
+//
+// Note: as this involves closing multiple iterators, only the first error
```
kortschak: See above.

barakmich:
@andrew-d That strange bit should really have a comment: 

```
// One of the iterators has determined that this value doesn't match. 
// However, the iterators that came before in the list may have returned "ok" to Contains(). 
// We need to set all the tags back to what the previous result was -- effectively 
// seeking back exactly one -- so we check all the prior iterators with the (already verified) result and 
// throw away the result, which will be 'true'
```

And check them for errors in that list. 

Well caught!

graph/iterator/hasa_iterator_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/google/cayley/quad"
+)
+
+func TestHasAIteratorErr(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: s/retErr/wantErr/g

graph/iterator/linksto_iterator.go
```diff
@@ -176,19 +189,38 @@ func (it *LinksTo) Next() bool {
...
 func (it *LinksTo) Result() graph.Value {
 	return it.result
 }
 
 // Close our subiterators.
-func (it *LinksTo) Close() {
-	it.nextIt.Close()
-	it.primaryIt.Close()
+//
+// Note: as this involves closing multiple subiterators, only the first error
```
kortschak: See above.

graph/iterator/materialize_iterator_test.go
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iterator
+
+import (
+	"errors"
+	"testing"
+
+	//"github.com/google/cayley/graph"
```
kortschak: Delete.

graph/iterator/materialize_iterator_test.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+func TestMaterializeIteratorError(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: s/retErr/wantErr/g

graph/iterator/materialize_iterator_test.go
```diff
@@ -0,0 +1,72 @@
...
+	if mIt.Next() != false {
+		t.Errorf("Materialize iterator did not pass through underlying 'false'")
+	}
+	if mIt.Err() != retErr {
+		t.Errorf("Materialize iterator did not pass through underlying Err")
+	}
+}
+
+func TestMaterializeIteratorErrorAbort(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: As above.

graph/iterator/materialize_iterator_test.go
```diff
@@ -0,0 +1,72 @@
...
+	// This tests that we properly return 0 results and the error when the
+	// underlying iterator is larger than our 'abort at' value, and then
+	// returns an error.
+	or := NewOr()
+	or.AddSubIterator(NewInt64(1, int64(abortMaterializeAt+1)))
+	or.AddSubIterator(errIt)
+
+	mIt := NewMaterialize(or)
+
+	// Should get all the underlying values...
```
kortschak: We should...

graph/iterator/materialize_iterator_test.go
```diff
@@ -0,0 +1,72 @@
...
+			t.Errorf("Materialize iterator returned spurious 'false' on iteration %d", i)
+			return
+		}
+		if mIt.Err() != nil {
+			t.Errorf("Materialize iterator returned non-nil Err on iteration %d", i)
+			return
+		}
+	}
+
+	// ... and then the error value
```
kortschak: Finish with a period.

graph/iterator/not_iterator.go
```diff
@@ -115,9 +130,19 @@ func (it *Not) NextPath() bool {
 	return false
 }
 
-func (it *Not) Close() {
-	it.primaryIt.Close()
-	it.allIt.Close()
+// Close closes the primary and all iterators.  If an error occurs, only the
```
kortschak: Use same phrasing as above.

graph/iterator/or_iterator.go
```diff
@@ -231,12 +252,21 @@ func (it *Or) cleanUp() {}
 
 // Close this graph.iterator, and, by extension, close the subiterators.
 // Close should be idempotent, and it follows that if it's subiterators
-// follow this contract, the And follows the contract.
-func (it *Or) Close() {
+// follow this contract, the Or follows the contract.
+//
+// Note: as this potentially involves closing multiple subiterators, only
```
kortschak: See above.

graph/iterator/or_iterator_test.go
```diff
@@ -148,3 +149,46 @@ func TestShortCircuitingOrBasics(t *testing.T) {
 		t.Errorf("Failed to iterate optimized Or correctly, got:%v expect:%v", got, expect)
 	}
 }
+
+func TestOrIteratorErr(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: s/retErr/wantErr/g

graph/iterator/or_iterator_test.go
```diff
@@ -148,3 +149,46 @@ func TestShortCircuitingOrBasics(t *testing.T) {
...
+	if or.Next() != false {
+		t.Errorf("Or iterator did not pass through underlying 'false'")
+	}
+	if or.Err() != retErr {
+		t.Errorf("Or iterator did not pass through underlying Err")
+	}
+}
+
+func TestShortCircuitOrIteratorErr(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: As above.

graph/iterator/value_comparison_iterator.go
```diff
@@ -168,7 +175,11 @@ func (it *Comparison) Contains(val graph.Value) bool {
 	if !it.doComparison(val) {
 		return false
 	}
-	return it.subIt.Contains(val)
+	ret := it.subIt.Contains(val)
```
kortschak: s/ret/ok/g

graph/iterator/value_comparison_iterator_test.go
```diff
@@ -118,3 +119,17 @@ func TestVCIContains(t *testing.T) {
 		}
 	}
 }
+
+func TestComparisonIteratorErr(t *testing.T) {
+	retErr := errors.New("unique")
```
kortschak: s/retErr/wantErr/g

graph/primarykey.go
```diff
@@ -84,8 +84,9 @@ func (p *PrimaryKey) Int() int64 {
 	case sequential:
 		return p.sequentialID
 	case unique:
-		glog.Fatal("UUID cannot be cast to an int64")
-		return -1
+		msg := "UUID cannot be cast to an int64"
```
kortschak: s/cast/converted/

writer/single.go
```diff
@@ -33,17 +33,24 @@ type Single struct {
 
 func NewSingleReplication(qs graph.QuadStore, opts graph.Options) (graph.QuadWriter, error) {
 	var ignoreMissing, ignoreDuplicate bool
+	var err error
```
kortschak: Use a var block.

kortschak:
I've done a first pass. In additions to the specific comments, three general comments:

1. Use `err` or `ok` (or other contextually meaningful name) in place of `ret`. To avoid shadowing where it happens because of this, rename temporary error values to something like `_err`.
2. Where you have `if expr; cond {` replace with `expr; if cond {` unless you really need the scoping or the readability if significantly improved (a good example of this is when you are doing a single map existance test), otherwise the conditional is harder to read.
3. Move err fields to bottom of the struct (this seems to be broadly conventional).

kortschak:
Oh, also, when this gets merged it should have "Updates #157." in the merge commit message.

andrew-d:
@barakmich, @kortschak: Okay, I think I addressed all your comments, and the message you mentioned is in the PR description (which I think means it gets into the merge commit).

graph/bolt/iterator.go
```diff
@@ -46,6 +46,7 @@ type Iterator struct {
 	dir     quad.Direction
 	qs      *QuadStore
 	result  *Token
+	err     error
```
kortschak: Move this one down.

graph/iterator/linksto_iterator.go
```diff
@@ -176,19 +187,34 @@ func (it *LinksTo) Next() bool {
...
+
 func (it *LinksTo) Result() graph.Value {
 	return it.result
 }
 
-// Close our subiterators.
-func (it *LinksTo) Close() {
-	it.nextIt.Close()
-	it.primaryIt.Close()
+// Close our subiterators.  It closes all subiterators it can, but
```
kortschak: Phrasing here.

andrew-d: Do you mind clarifying?  I'm not seeing anything wrong here, but I might be missing something.

kortschak:
That style is better thanks.

LGTM with minor nits.

graph/iterator/linksto_iterator.go
kortschak: Sorry, clarified in edit.

andrew-d:
Okay, I've addressed all @kortschak's comments :smile:

Let me know if anything else needs to be done!

barakmich:
Just cause I've gotta ask: You've signed the CLA?

andrew-d:
If you mean [this](https://cla.developers.google.com/), I have now :smile_cat:

barakmich:
Excellent! Merging this, as it LGTM too with another look.

A good followup PR is adding yourself to the A+C, @andrew-d . :+1:

PR #233, opened 2015-04-16
fix decompressing for http

barakmich:
Move the decompression detection to quad (it seeming to be the most logical place for it) and support uploading compressed files through http.

### DIFF
--- a/cayley.go
+++ b/cayley.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/cayley/db"
 	"github.com/google/cayley/graph"
 	"github.com/google/cayley/http"
+	"github.com/google/cayley/internal"
 	"github.com/google/cayley/quad"
 	"github.com/google/cayley/quad/cquads"
 	"github.com/google/cayley/quad/nquads"
@@ -286,7 +287,7 @@ func decompressAndLoad(qw graph.QuadWriter, cfg *config.Config, path, typ string
 		r = res.Body
 	}
 
-	r, err = quad.Decompressor(r)
+	r, err = internal.Decompressor(r)
 	if err != nil {
 		if err == io.EOF {
 			return nil
--- a/cayley.go
+++ b/cayley.go
@@ -17,10 +17,6 @@
 package main
 
 import (
-	"bufio"
-	"bytes"
-	"compress/bzip2"
-	"compress/gzip"
 	"flag"
 	"fmt"
 	"io"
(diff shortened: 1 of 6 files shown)


proteneer:
Does this deal with zlib as well? AFAIK zlib has a different magic header from gzip, and has caused me tons of pain in the past.

kortschak:
I don't think this is the right thing to do; decompression and quad handling are distinct concerns. If the decompression sniffer is factored out, I would rather see it moved to a cayley/internal package.

barakmich:
SGTM. zlib detection is a great small addition to this, if you'd like to give it a go @proteneer

kortschak:
Move cayley.TestDecompressor and its associated var into internal too I think.

kortschak:
LGTM

kortschak:
@proteneer the change to add zlib is simple, but subtle.

The following snippet shows what is sniffed as zlib but is also valid ASCII text (and valid cquads input - nquads is safer here - because cquad grammar [allows string literals](https://github.com/google/cayley/blob/master/quad/cquads/cquads.rl#L55)):

http://play.golang.org/p/_-9ZfbPX9w

(it is actually worse than this because quads is defined to parse utf-8 unicode¸ so j in the inner loop is less constrained).

We already have this problem with bzip2 (the sniffer precludes cquad data with a "BZh" prefix), but it would be good to not compound that too much.

If you want to do this, you will need to add a helper that reflects code in compress/zlib.reader.Reset. You will also need to extend the peeked bytes to 6 bytes from the current 3.

```
const zlibDeflate = 8

func isZlib(buf []byte) bool {
	h := uint(buf[0])<<8 | uint(buf[1])
	if (buf[0]&0x0f != zlibDeflate) || (h%31 != 0) {
		return false
	}
	haveDict := buf[1]&0x20 != 0
	if haveDict {
		checksum := uint32(buf[2])<<24 | uint32(buf[3])<<16 | uint32(buf[4])<<8 | uint32(buf[5])
		if checksum != 1 /*adler32.Checksum(nil)*/ {
			return false
		}
	}
	return true
}
```

This still gives us 576 matches for all possible 6-bytes. This is reasonably safe since most are not legal cquad (none are valid nquad) syntax - 6 valid cquad ASCII digraph prefixes are still returned as zlib streams, {"(S", "8O", "HK", "XG", "hC", "x^"} (18 utf-8 unicode prefixes).

PR #234, opened 2015-04-16
Vendor dependencies with Godep

jzelinskie:
This change introduces the usage of Godep in order to vendor dependencies and create reproducible builds.


kortschak:
With the view that cayley will be offering itself as a Go API in addition to the current mql/gremlin APIs, and so included as an import outside the cayley tree: this NOT LGTM.

Please see [discussion at golang-dev](https://groups.google.com/d/topic/golang-dev/nMWoEAG55v8/discussion) on the conventions behind vendoring in libraries: [adg](https://groups.google.com/d/msg/golang-dev/nMWoEAG55v8/GXh7DT7HknoJ) and [rsc](https://groups.google.com/d/msg/golang-dev/nMWoEAG55v8/GXh7DT7HknoJ).

If you can vendor only the cayley app dependencies then it would be OK, but I don't see this as being all that useful since the majority of external are being imported in the non-main packages.

jzelinskie:
I understand your point, but there is also something to be said of not being able to reproduce builds.
How about the compromise of not rewriting imports? Then the use of Godep becomes optional, but we can require that all _releases_ of Cayley have their Godeps synchronized such that the build can be reproduced by checking out the tag of the release and then running `godep go build`.

kortschak:
I'd be happier if we weren't keeping all the source if imports are not being rewritten anyway. What about glock? This is used in cockroach.

jzelinskie:
I'm not familiar with glock, but godep also supports this behavior if you commit only the `Godeps/Godeps.json` file and use `godep get` to download those versions of your dependencies into your $GOPATH.

kortschak:
OK SGTM

jzelinskie:
Ok, I'm going to close this and open a new PR then.

paultag:
:+1: btw; forgot to comment, but not ACKing this is a huge step up. Thanks for not vendoring!

PR #235, opened 2015-04-17
store dependencies' versions with godep

jzelinskie:
These dependencies can be loaded into your $GOPATH simply executing
`godep get`. This will allow builds to become reproducible, without
getting in the way of those wanting to consume parts of the project's
source code as a library.

As per the discussion in #234.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,23 +8,9 @@ go:
   - tip
 
 install:
-  - go get github.com/badgerodon/peg
-  - go get github.com/barakmich/glog
-  - go get github.com/cznic/mathutil
-  - go get github.com/julienschmidt/httprouter
-  - go get github.com/petar/GoLLRB/llrb
-  - go get github.com/peterh/liner
-  - go get github.com/robertkrimen/otto
-  - go get github.com/russross/blackfriday
-  - go get github.com/syndtr/goleveldb/leveldb
-  - go get github.com/syndtr/goleveldb/leveldb/cache
-  - go get github.com/syndtr/goleveldb/leveldb/iterator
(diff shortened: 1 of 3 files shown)


jzelinskie:
If this sounds good to everyone, I can go ahead and update the .travis.yml to use this method to create builds, this way we can know that the dependencies are always maintained.

kortschak:
LGTM

barakmich:
Yeah. Given the Cayley-as-a-library notion, this is probably the right net answer. `.travis.yml` support is probably a good idea too!

jzelinskie:
Looks like `godep get` is restoring all but the mathutil package. I'll continue to look into it tomorrow, as it's gotten pretty late.

jzelinskie:
I really should also write something for the docs, as changing a dependency requires executing `godep save --copy=false ./...` and loading the vendored dependencies in your $GOPATH requires executing `godep restore`

jzelinskie:
Looks like godep was hanging on the tip build; likely related to https://github.com/tools/godep/issues/206

kortschak:
Note also this recent post https://groups.google.com/d/topic/golang-nuts/G0HO7sEsIQQ/discussion

Maybe we could hold off until a canonical dep file format is actually decided in the nuts thread I linked #234.

jzelinskie:
If it wasn't for there being an issue with tip, I would argue that having this merged now hurts no one and only gives us the benefit of having reproducible builds until the Go team figures out what they want to do.

I don't know how long it's going to take them, but I hope it's sooner rather than later.

jzelinskie:
Rebased into one commit. The issue with tip seems to be resolved now.

kortschak:
LGTM

PR #237, opened 2015-04-21
added example n-quad file for gremlin api description

jf87:
Just felt it was useful to have the example graph used in https://github.com/google/cayley/blob/master/docs/GremlinAPI.md to try out the described queries by loading the graph into cayley.


barakmich:
Good idea, but I think we should push it slightly to be more awesome.

1) Perhaps it's time to have a `/data` directory, with some of the example datafiles (that are currently in root), this one included.
2) This file (sort of) exists as `query/gremlin/gremlin_test.nt` -- yours should become canonical and the gremlin test should be modified slightly to use it (changing cool to cool_person and the load path in the gremlin test)
3) Add a few lines to GremlinAPI.md on "play around with this simple graph using the command...."
4) All documentation referring to datasets in root now refer to them in the `data` dir.

This also opens the door for other data sets, some data set integration/benchmark tests and so on. In short, really useful :)

jf87:
Sounds good :-)
I could do this during the next days and add it to this pull request.

barakmich:
Awesome. Totally a great first PR with some improvements in a few corners :)

jf87:
I've added the above changes now. Let me know what you think :-)

docs/GremlinAPI.md
```diff
@@ -163,9 +169,9 @@ Filter all paths to ones which, at this point, are on the given `node`.
 
 Example:
 ```javascript
-// Starting from all nodes in the graph, find the paths that follow B.
-// Results in three paths for B (from A C and D)
-g.V().Out("follows").Is("B")
+// Starting from all nodes in the graph, find the paths that follow bob.
+// Results in three paths for bob (from alic, charlie and dani)
```
barakmich: alice

barakmich:
LGTM. Merging. 

We were super impressed in the office that you went to the trouble to correct the ASCII art :)

jf87:
Haha, I had already started to change the names when I remembered the ASCII graph, so I had to go through with it as well :-)

PR #242, opened 2015-04-26
Implement the Unique iterator

barakmich:
Uniqueness is a feature request people have asked for before -- returning the first path to some node in the middle of a query tree.


allonhadaya:
LGTM :+1:

PR #243, opened 2015-04-26
In out predicates

### LINKED ISSUE #239: Add Predicate/Edge Type functionality to query language
There should be a way to query the graph to return all predicates/edge types so that we know what can be used in the In( ) and Out( ) commands. For example, if the 30kmoviedata.nq.gz data, it would be nice to know if something like g.V( ).Out("director").All( ) would fail or not without having to go through trial and error. Ideally, something like g.predicates( ).All( ) would return a list of the nodes in the graph with a "relationship" type. 

The predicate function should allow the use of both .All( ) and .GetLimit( ) so the user can limit the number of values returned if needed.

allonhadaya:
Fixes #239

### DIFF
--- a/query/gremlin/build_iterator.go
+++ b/query/gremlin/build_iterator.go
@@ -139,6 +139,15 @@ func buildInOutIterator(obj *otto.Object, qs graph.QuadStore, base graph.Iterato
 	return iterator.NewHasA(qs, and, out)
 }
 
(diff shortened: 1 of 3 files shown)


allonhadaya:
I've signed the CLA.

barakmich:
LGTM

PR #247, opened 2015-05-01
Clean up Linkage definition

barakmich:
A linkage is a direction/value pair, and that's all. A set of these can
have helper functions and the like. Cleans up some assumptions of
functionality that isn't useful (yet).

Refactor of a new introduction.

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -38,9 +38,11 @@ type Tagger struct {
 // Linkage is a union type representing a set of values established for a given
 // quad direction.
 type Linkage struct {
-	Dir    quad.Direction
-	Values []Value
+	Dir   quad.Direction
+	Value Value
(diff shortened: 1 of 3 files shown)


graph/iterator.go
```diff
@@ -39,9 +39,14 @@ type Tagger struct {
 // quad direction.
 type Linkage struct {
 	Dir    quad.Direction
-	Values []Value
+	Value Value
 }
 
+// LinkageSet is an array of Linkages with some helpful functions for aggregation.
+type LinkageSet []Linkage
```
kortschak: Linkages? - Set suggests unordered. Up to you.

kortschak:
LGTM

Maybe only add LinkageSet when []Linkage needs methods.

PR #255, opened 2015-05-22
internal: move {config,db,http} into internal

kortschak:
This is a straw man, but the intention is that if we move these packages into internal, we are still free to alter them without breaking API compatibility. Initially I was just thinking http, but the others make sense as well.

If we want to publish these we can pull them out as needed at a later date.


barakmich:
I like the notion, but I'm thinking that http is kind of already published, in its own way; even though we're still pre-1.0. The other two seem correct though.

The full idea being, of course, that a library need not touch these things. If that's the intent, then sure; then maybe that's what internal means (or maybe `server` or the like, to capture publishing?)

kortschak:
The intention is that only cmd/cayley should actually need to use cayley/http by whatever name. I doubt that anyone else imports it since it is very cayley specific, but this is to capture that state so that it can be improved more freely.

On 30/05/2015, at 5:04 AM, "Barak Michener" <notifications@github.com<mailto:notifications@github.com>> wrote:


I like the notion, but I'm thinking that http is kind of already published, in its own way; even though we're still pre-1.0. The other two seem correct though.

The full idea being, of course, that a library need not touch these things. If that's the intent, then sure; then maybe that's what internal means (or maybe server or the like, to capture publishing?)

kortschak:
Sorry, are you suggesting http => server? If so, I'm good with that.

kortschak:
@barakmich PTAL

kortschak:
@barakmich *ping*

.travis.yml
```diff
@@ -14,8 +14,8 @@ install:
 
   # Google App Engine dependencies
   - cd ..
-  - wget https://storage.googleapis.com/appengine-sdks/featured/go_appengine_sdk_linux_amd64-1.9.19.zip
-  - unzip -q go_appengine_sdk_linux_amd64-1.9.19.zip
+  - wget https://storage.googleapis.com/appengine-sdks/featured/go_appengine_sdk_linux_amd64-1.9.21.zip
```
jzelinskie: You should rebase this, because this is fixed upstream.

kortschak: Done.

kortschak:
@barakmich I'm trying to understand what is blocking this. I'm not getting any feedback on why this is not accepted. It conflicts again, but I'm not sure whether it is worth my time to keep resolving conflicts if the PR is just going to dangle here.

None of the packages that are being moved in this PR were used by packages except cmd/cayley and leaving them public risks that changing, thus preventing future modifications (which IMO need to happen, at least for the http server).

barakmich:
That's on me. I'm happy to hand-merge it. I just lost the thread. LGTM and immediate merge if you're up for a rebase, otherwise I'll do it :)

kortschak:
Rebase is done and push -f'd.

Thanks.

PR #258, opened 2015-05-28
Fix nil pointer panic when meta bucket is missing

op:
If `NewQuadStore` was called without first calling `InitQuadStore`, the runtime would panic since the meta bucket never was created.

Feel free to close this PR and do it differently. Ran into this bug when I explored cayley for the first time. Hopefully it saves more gopher puppies from running into a nil pointer panic.

I signed the agreement in case.

### DIFF
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -37,6 +37,10 @@ func init() {
 }
 
 var (
+	errNoBucket = errors.New("bolt: bucket is missing")
+)
+
+var (
 	hashPool = sync.Pool{
 		New: func() interface{} { return sha1.New() },
 	}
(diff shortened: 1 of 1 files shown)


kortschak:
s/errBucketNotExist/errNoBucket/g but it seems to me that if you call NewQuadStore without having done the init, a panic is appropriate, maybe just with a better message and not just a nil panic.

op:
Alright. Fixed!

graph/bolt/quadstore.go
```diff
@@ -97,7 +101,9 @@ func newQuadStore(path string, options graph.Options) (graph.QuadStore, error) {
 		return nil, err
 	}
 	err = qs.getMetadata()
-	if err != nil {
+	if err == errNoBucket {
+		panic("bolt: quad has not been initialised")
```
kortschak: s/quad/quadstore/

barakmich:
LGTM.

PR #260, opened 2015-05-28
collapse 2 iterations into 1 in memstore

neonstalwart:
i'm reading the code to try to comprehend how everything is implemented and it looked like these 2 iterations could be collapsed into 1.  let me know if i've missed something subtle.


barakmich:
Yeah, I think it's fine. There may have been a subtlety there at one time (and I think there was, with the creation of something conditionally) but I can't see anything crazy here. Plus, we'd know pretty straightaway with the tests.

LGTM

PR #262, opened 2015-05-29
add string comparison to value comparison iterator

neonstalwart:
i was reading the code and saw a TODO to add string comparison to the value comparison iterator.  i thought it was within my reach to tackle it so here's my attempt.  let me know what you think.


barakmich:
LGTM

As a next step, a way of exposing the iterator in the query language would be kind of cool :)

PR #264, opened 2015-05-29
simplify the signature of NextLogOut

neonstalwart:
by getting the value from `it.Result()` rather than passing it directly, this
reduces the number of things which need to be passed and may help identify
cases where the value was passed to `NextLogOut` but was not set in a way to
make `it.Result()` work as it should. in those cases, the logging would
include a result different to what you would get from `it.Result()` but lead
you to believe that `it.Result()` would return the logged value.

this is based on seeing a lot of this pattern repeated:

```go
graph.NextLogOut(it, nil, false)
  ...
it.result = value
graph.NextLogOut(it, value, true)
```

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -101,7 +101,7 @@ type Iterator interface {
 	//  	}
 	//  }
 	//
-	// All of them should set iterator.Last to be the last returned value, to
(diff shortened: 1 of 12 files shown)


dennwc:
It took a long time, but now it's merged. Thanks for the PR!

PR #265, opened 2015-05-30
Allow opening a database (bolt, leveldb, mongo) through Go API

Quentin-M:
This small PR allows opening a database through Go code (#208)

**Usage Example**
```
package main

import "github.com/google/cayley"

store, err := cayley.NewGraph("bolt", "/home/mentalow/mexican_coke_db", nil)
if err != nil {
	log.Fatalln(err)
}
```

(CLA Accepted)

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -28,8 +28,8 @@ func Quad(subject, predicate, object, label string) quad.Quad {
 	return quad.Quad{subject, predicate, object, label}
 }
 
-func NewMemoryGraph() (*Handle, error) {
-	qs, err := graph.NewQuadStore("memstore", "", nil)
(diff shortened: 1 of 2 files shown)


imports.go
```diff
@@ -2,7 +2,10 @@ package cayley
 
 import (
 	"github.com/google/cayley/graph"
+	_ "github.com/google/cayley/graph/bolt"
```
barakmich: I'd rather let the user choose the imports for non-memory graphs for their use case. Beyond that, the function below looks fine.

jzelinskie:
looks good if you flatten the commits :+1:

Quentin-M:
Cleaned history.

PR #268, opened 2015-06-02
Allow opening a database (bolt, leveldb, mongo) through Go API

Quentin-M:
This small PR allows opening a database through Go code (#208)

**Usage Example**
```
package main

import "github.com/google/cayley"

store, err := cayley.NewGraph("bolt", "/home/mentalow/mexican_coke_db", nil)
if err != nil {
	log.Fatalln(err)
}
```

(CLA Accepted)

### DIFF
--- a/imports.go
+++ b/imports.go
@@ -28,8 +28,8 @@ func Quad(subject, predicate, object, label string) quad.Quad {
 	return quad.Quad{subject, predicate, object, label}
 }
 
-func NewMemoryGraph() (*Handle, error) {
-	qs, err := graph.NewQuadStore("memstore", "", nil)
+func NewGraph(name, dbpath string, opts graph.Options) (*Handle, error) {
+	qs, err := graph.NewQuadStore(name, dbpath, opts)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


barakmich:
LGTM

kortschak:
I don't like that this is stringly-typed. It made sense when we were do things for a stringly-typed front end but here we have a properly typed™ language as the front end. I think we can improve this by defining and exposing a pair of types that do the work of registration:

```
type Registerer interface {
        NewQuadStore() (graph.QuadStore, error)
}

type DBCreator interface {
        CreateDB(path string, options graph.Options) (graph.QuadStore, error)
}
```

All quadstore packages must provide a Registerer which may also implement DBCreator. The implementation need only be a struct{} with the current function pinned on as methods.

With this we can then say:

```
package main

import "github.com/google/cayley"

store, err := cayley.NewGraph(bolt.Store{}, "/home/mentalow/mexican_coke_db", nil)
if err != nil {
        log.Fatalln(err)
}
```

barakmich:
To be fair to the PR, there's certainly precedent for this style in database/sql.

Whether there's a better way is a perfectly cromulent issue :)

kortschak:
I'm happy to propose this, it's a trivial change.

barakmich:
The registration interface might also wrap the type -- `QuadStoreName() string` or equivalent, as a superset of what exists today.

kortschak:
That seems fair.

PR #270, opened 2015-06-05
Small REPL Improvements

tmlbl:
Forgive me if these are naive, but here are a couple small modifications I made to the REPL.


barakmich:
Yeah, sure.

PR #271, opened 2015-06-10
Widened visualisation box to 100%

nii236:
It was very frustrating to use the visualisation feature of Cayley, as the vis box was tiny and waste a lot of space. I widened the space to 100% of the viewable DOM. This makes it much more pleasant to use.

Before:
screen shot 2015-06-10 at 16 10 43
After:
screen shot 2015-06-10 at 16 10 31

I have signed the Google CLA.

### DIFF
--- a/static/css/query_editor.css
+++ b/static/css/query_editor.css
@@ -40,6 +40,6 @@ line.link {
 }
 
 #visualize {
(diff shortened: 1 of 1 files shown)


barakmich:
Awesome! Thank you.

PR #274, opened 2015-06-15
Add transaction (a list of deltas) as a primitive for applying sets o…

barakmich:
…f changes

This will also line up with removing "AddQuadSet" and help with some more intricate types of updates. It's not full-query-capability yet, but we'll need the primitive to do that anyway.

### DIFF
--- a/graph/transaction.go
+++ b/graph/transaction.go
@@ -1,0 +1,43 @@
+// Copyright 2015 The Cayley Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


graph/transaction.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"github.com/google/cayley/quad"
+)
+
+type Transaction struct {
+	Deltas []Delta
+}
+
+func CreateTransaction() *Transaction {
```
jzelinskie: `NewTransaction` would be more idiomatic

imports.go
```diff
@@ -18,6 +18,7 @@ var StartMorphism = path.StartMorphism
 var StartPath = path.StartPath
 
 var RawNext = graph.Next
+var CreateTransaction = graph.CreateTransaction
```
jzelinskie: move these exports into a single `var` block

jzelinskie:
LGTM

Quentin-M:
LGTM

PR #278, opened 2015-06-26
Use new import path for go-uuid

schmichael:
code.google.com is going away


barakmich:
This is correct; merging (failure is because appengine, merging #282 simultaneously)

PR #279, opened 2015-07-03
updating UUID package dependency

hiteshjoshi:
The go-UUID package has been moved from https://code.google.com/p/go-uuid/ to github.


barakmich:
Given that we don't need the vendoring checked in, has an odd `root@` binary, and #278 uses the same SHA, I'm going with #278 instead -- but thanks for dealing with `godep` (which I know is a pain :( )

hiteshjoshi:
No problem. What's odd about it though?

PR #287, opened 2015-07-28
Fixed multiplier for position of "label" in case of "po" key entry

igm:
in "po", the ordering is {"predicate","object","subject","label"} thus the offset of label is 3*hashSize+2


barakmich:
Good catch.

PR #289, opened 2015-07-31
graph: Postgres backend

### LINKED ISSUE #25: Postgres backend!
One of the things in __[TODO.md](https://github.com/google/cayley/blob/bbb0a2f5807b0b6a33dee71e7ef9ccedf4fdc050/TODO.md "cayley/TODO.md at bbb0a2f5807b0b6a33dee71e7ef9ccedf4fdc050 · google/cayley")__ is supporting Postgres as a backend. As someone who loves Postgres when dealing with relational data, I'd like to hear more thoughts on this, because I'd be interested in implementing it. For example, would native data types be best or could every entry just make use of Postgres' awesome native support for JSON documents?

As stated in the document, *why not* :smile:?

\- Jonathan

__Note:__ I've signed the CLA.

barakmich:
Fixes #25 

So we've been exploring a Postgres Cayley backend at Quay, and after about 3 or 4 iterations on the implementation, I think I've finally got one that works fairly well.

The idea is this -- an SQLIterator follows the graph.Iterator interface and knows how to build a query, based on an (internal, private) sqlIterator -- which may build its own tree and actually builds the SQL (many iterations on making the query performant were run).

To wit, for a number of standard query cases, a single SQL query can be generated and run fairly fast, given appropriate (automatically created) indexes on the quads table. Some baseline speeds (which can likely be improved):

```
PASS
BenchmarkNamePredicate	    1000	   1859128 ns/op
BenchmarkLargeSetsNoIntersection	     100	  18201976 ns/op
BenchmarkVeryLargeSetsSmallIntersection	      50	  37389066 ns/op
BenchmarkHelplessContainsChecker	     300	   5557484 ns/op
BenchmarkHelplessNotContainsFilms	     100	  10980546 ns/op
BenchmarkHelplessNotContainsActors	       1	4248740907 ns/op
BenchmarkNetAndSpeed	      10	 179705838 ns/op
BenchmarkKeanuAndNet	      20	  95033993 ns/op
BenchmarkKeanuAndSpeed	      20	  85134634 ns/op
BenchmarkKeanuOther	      10	 175570195 ns/op
BenchmarkKeanuBullockOther	       3	 346969106 ns/op
ok  	github.com/google/cayley/integration	119.134s
```

So using Postgres as a graph isn't too shabby (single-to-tens-of-ms for some standard sorts of queries. Postgres is cool.)

In cases of trees that can't be collapsed to a single query, it does as much as it can and then runs the query the old-fashioned way with more roundtrips. So while nothing is lost, slowdowns for unoptimized SQL options are possible extensions. Making the SQL creation extensible was tricky -- it's still a little rough, but at least its somewhat sane.

Todo before merge:
- [ ] Docs


cmd/cayley/cayley.go
```diff
@@ -147,6 +149,15 @@ func main() {
...
 	flag.Parse()
 
+	if *cpuprofile != "" {
+		f, err := os.Create(*cpuprofile)
+		if err != nil {
+			glog.Fatal(err)
+		}
+		pprof.StartCPUProfile(f)
+		defer pprof.StopCPUProfile()
+	}
```
jzelinskie: I've definitely never written this before... ;)

graph/sql/quadstore.go
```diff
@@ -0,0 +1,281 @@
...
+			//for _, dir := range quad.Directions {
+			//_, err := tx.Exec(`
+			//WITH upsert AS (UPDATE nodes SET size=size+1 WHERE node=$1 RETURNING *)
+			//INSERT INTO nodes (node, size) SELECT $1, 1 WHERE NOT EXISTS (SELECT * FROM UPSERT);
+			//`, d.Quad.Get(dir))
+			//if err != nil {
+			//glog.Errorf("couldn't prepare upsert statement in direction %s: %v", dir, err)
+			//return err
+			//}
+			//}
```
jzelinskie: Should replace this with a TODO

graph/sql/quadstore.go
```diff
@@ -0,0 +1,281 @@
...
+		case graph.Delete:
+			_, err := tx.Exec(`DELETE FROM quads WHERE subject=$1 and predicate=$2 and object=$3 and label=$4;`,
+				d.Quad.Subject, d.Quad.Predicate, d.Quad.Object, d.Quad.Label)
+			if err != nil {
+				glog.Errorf("couldn't prepare DELETE statement: %v", err)
+			}
+			//for _, dir := range quad.Directions {
+			//tx.Exec(`UPDATE nodes SET size=size-1 WHERE node=$1;`, d.Quad.Get(dir))
+			//}
```
jzelinskie: Another commented out block.

barakmich:
Seeing as we've been building prod on it, I'm happy to call this feature built. Bugs will exist, of course, but let's make it master.

saidimu:
@barakmich Big props on this feature!

Caveats on usage?

barakmich:
@saidimu : Works pretty well over here. The biggest caveats are unoptimized cases (like #316) that require multiple roundtrips.

PR #291, opened 2015-08-04
Update README.md

skynet:
Fix link to Google's [Knowledge Graph](https://www.google.com/search/about/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <p align="center">
   <img src="static/branding/cayley_side.png?raw=true" alt="Cayley" />
 </p>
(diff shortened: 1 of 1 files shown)


jzelinskie:
I doesn't look like that page has anything to do with Knowledge Graph anymore. We should probably just link to Wikipedia: https://en.wikipedia.org/wiki/Knowledge_Graph

skynet:
:+1: Quite strange not to see a dedicated place at Google for the Fix link to [Knowledge Graph](https://en.wikipedia.org/wiki/Knowledge_Graph).

Issue #176, opened 2014-10-17
API for exporting graph data


saidimu:
The only way right now is to iterate the database and save the data client-side.

An export API which returns a gzipped file would be more efficient. Even better if a variety of [export formats](http://gephi.github.io/users/supported-graph-formats/) are supported.

The primary motivation for this is to be able to pair Cayley with a visualization and exploration platform like [Gephi](https://gephi.github.io/).

PR #292, opened 2015-08-10
Adds exporting capabilities

ds--:
My first attempt at writing/modifying go code, so forgive possible egregious idiomatic golang mistakes. So here it is.

PR Adds exporting capabilities for backup/migration/visualization

    ./cayley dump -config=foo.cfg -dump=backup.nq -dump_type=nquad

**Types:** nquad / json / [gml](https://en.wikipedia.org/wiki/Graph_Modelling_Language) / [graphml](https://en.wikipedia.org/wiki/GraphML) (last two experimental)
**Output:** to file, gzipped file (when .gz extension used), or "-" for stdout

Solves #176 too.

### DIFF
--- a/cmd/cayley/cayley.go
+++ b/cmd/cayley/cayley.go
@@ -50,7 +50,7 @@ var (
 	databasePath       = flag.String("dbpath", "/tmp/testdb", "Path to the database.")
 	databaseBackend    = flag.String("db", "memstore", "Database Backend.")
 	dumpFile           = flag.String("dump", "dbdump.nq", `Quad file to dump the database to (".gz" supported, "-" for stdout).`)
-	dumpType           = flag.String("dump_type", "json", `Quad file format ("json", "nquad").`)
+	dumpType           = flag.String("dump_type", "quad", `Quad file format ("json", "quad", "gml", "graphml").`)
 	replicationBackend = flag.String("replication", "single", "Replication method.")
 	host               = flag.String("host", "127.0.0.1", "Host to listen on (defaults to all).")
 	loadSize           = flag.Int("load_size", 10000, "Size of quadsets to load")
--- a/cmd/cayley/cayley.go
+++ b/cmd/cayley/cayley.go
@@ -219,7 +219,11 @@ func main() {
 			}
 		}
 
-		// internal.Dump()
+		err = internal.Dump(handle.QuadStore, *dumpFile, *dumpType)
+		if err != nil {
+			break
+		}
+
 		handle.Close()
 
 	case "repl":
--- a/cmd/cayley/cayley.go
+++ b/cmd/cayley/cayley.go
@@ -49,6 +49,8 @@ var (
 	configFile         = flag.String("config", "", "Path to an explicit configuration file.")
 	databasePath       = flag.String("dbpath", "/tmp/testdb", "Path to the database.")
 	databaseBackend    = flag.String("db", "memstore", "Database Backend.")
+	dumpFile           = flag.String("dump", "dbdump.nq", `Quad file to dump the database to (".gz" supported, "-" for stdout).`)
+	dumpType           = flag.String("dump_type", "json", `Quad file format ("json", "nquad").`)
 	replicationBackend = flag.String("replication", "single", "Replication method.")
 	host               = flag.String("host", "127.0.0.1", "Host to listen on (defaults to all).")
 	loadSize           = flag.Int("load_size", 10000, "Size of quadsets to load")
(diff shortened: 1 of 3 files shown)


cmd/cayley/cayley.go
```diff
@@ -204,6 +207,25 @@ func main() {
 
 		handle.Close()
 
+	case "dump":
+		handle, err = db.Open(cfg)
+		if err != nil {
+			break
+		}
+		if !graph.IsPersistent(cfg.DatabaseType) {
```
barakmich: The irony here is you're about to load a graph to dump a graph.

This is not wrong, however! It's using Cayley as a sort of graph-centered imagemagick-`convert`-esque. :+1:

exporter/exporter.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+type Exporter struct {
+	wr io.Writer
+	qstore graph.QuadStore
+	err error
+	count int32
+}
+
+func NewExporter(writer io.Writer, qstore graph.QuadStore) *Exporter {
```
barakmich: One thought would be to have a `NewExporterForIterator` -- given a quad-`graph.Iterator` -- so that applications might dump subgraphs. This functions would fill this in with `qstore.QuadsAllIterator()` as below.

Quad iterators exist (you found one :) ) but are hard to generate in the query languages -- for now, at least. But they could still be done by hand/in the future.

Taking quad iterators is perfectly sensible for the exporter package, IMO

barakmich:
This is shockingly good for a first writing/modifying attempt at Go. (@jzelinskie gave it an idiomatic look-over too, earlier) Well done! Just some slight modifications.

Any future goals or parts you'd be interested in? Cause it'd be great to have another pair of hands, on any part that might catch your eye.

Very related to this is an expanded delete API (instead of one quad, taking a quad iterator...) and, if you're interested in visualization, there's always other parsers (importing from graphml, say) -- though the latter is bound to be a bit more tricky, as XML is a large spec. But there is a built-in Go parser, so aiming for a subset of importable graphs (gephi, et al?) may not be too bad.

cmd/cayley/cayley.go
ds--: That was actually the very intention I had in mind. This is just food for thought but in the future it could be expended so something along those lines would be possible:

```./cayley dump -db="memstore" -quads="data.nq" -dump="subset.nq" -dump_type="nquad" -dump_query="g.V("charlie").Out("follows")"```

Not sure how much that's actually useful in practice but I could definitely see it come in handy on various occasions.

File: cmd/cayley/cayley.go [hunk 3]
```diff
@@ -49,6 +49,8 @@ var (
 	configFile         = flag.String("config", "", "Path to an explicit configuration file.")
 	databasePath       = flag.String("dbpath", "/tmp/testdb", "Path to the database.")
 	databaseBackend    = flag.String("db", "memstore", "Database Backend.")
+	dumpFile           = flag.String("dump", "dbdump.nq", `Quad file to dump the database to (".gz" supported, "-" for stdout).`)
+	dumpType           = flag.String("dump_type", "json", `Quad file format ("json", "nquad").`)
```
kortschak: I would call this "quad" rather than nquad since we don't validate that the output is legal nquad.

exporter/exporter.go
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+	"io"
+	"encoding/json"
+	"strconv"
+
+	"github.com/google/cayley/graph"
+)
+
+type Exporter struct {
```
kortschak: This really looks to me like a job for an interface with a variety of marshalers. Look to [encoding.TextMarshaler](http://golang.org/pkg/encoding/#TextMarshaler) for how you might do this.

Going with a collection of registered interface implementation eases extensibility.

exporter/exporter.go
```diff
@@ -0,0 +1,180 @@
...
+	wr io.Writer
+	qstore graph.QuadStore
+	err error
+	count int32
```
kortschak: Why is this int32?

exporter/exporter.go
```diff
@@ -0,0 +1,180 @@
...
+			return
+		}
+		exp.Write(string(jstr[:]))
+	}
+	exp.Write("]\n")
+}
+
+//experimental
+func (exp *Exporter) ExportGml() {
+	var seen map[string]int32 // todo eliminate this for large dbs
```
kortschak: Using a map[string]struct{} would save space here.

kortschak:
It's worth running gofmt on this.

barakmich:
LGTM

PR #293, opened 2015-08-12
Add check for multiple all iterators

barakmich:
If all the subiterators of an And iterator are Alls, we can return the first one. This hinders other potential optimizations (collapsing of queries, perhaps)


Quentin-M:
Cherry-picked and tested. Well done!

LGTM

PR #303, opened 2015-09-02
Fixes memstore transaction semantics

barakmich:


graph/memstore/quadstore.go
```diff
@@ -120,7 +140,7 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 				err = nil
 			}
 		default:
-			err = errors.New("memstore: invalid action")
+			panic("memstore: still invalid action?")
```
kortschak: This should never happen, but s/still invalid action\?/unexpected invalid action/

kortschak:
@Quentin-M the logic here looks like the correct way to do things, so those cases should be fixed to match this.

LGTM with minor comment.

barakmich:
@Quentin-M @kortschak -- correct. Those cases may have opportunity to be caught in the writer interface, however.

Fixed the nit, merging.

PR #308, opened 2015-09-15
vendor dependencies using go1.5 vendor experiment

azylman:
This means you can't build cayley without go1.5, but also means that _can_ build cayley without godep.


dennwc:
It seems like not a good idea to pull all dependencies source code into the repo. So closing this now.

PR #312, opened 2015-10-02
Refactor for readability

Kavec:
```
CHANGES: 
Added more comments, converted logical operations to (hopefully)
consistent 'route' metaphor, and moved all AND(iterator, iterator)
stanzas out into a single join() function that is more descriptive of
what's actually happening
```

Over the past few months, I've been working on a side-project involving some very extensive changes to Cayley to make it nicer to work with as a Golang library. Notably, after deleting 35k lines of *stuff*, I had to add fields to the Quad structure<sup>1</sup>.  That particular project is on hold for the moment, so I can now port<sup>2</sup> back improvements from that project to Cayley.

This section of Cayley, in particular, boggled me for days when I first read it-- this was prior to @barakmich's talk on the database internals and it took tracing execution of several iterator trees during debugging sessions to actually grok how the logical operators interacted. Hopefully these changes make it easier to parse and understand, encouraging more work and extension of the Cayley code.

Please let me know if there are any stylistic or technical issues in here before I drop pull requests for more substantial work (new Cassandra driver, caching system, In/Out morphism tagging, functions vice tags in the golang API, and more!).

(I have signed the Google CLA)

1. The comments in quad/quad.go aren't kidding, that's a rough experience
2. And it is, in fact, porting and translation at this point (For instance, 'quad' stopped being a properly descriptive name...)

### DIFF
--- a/graph/path/morphism_apply_functions.go
+++ b/graph/path/morphism_apply_functions.go
@@ -57,7 +57,8 @@ func isMorphism(nodes ...string) morphism {
 	}
 }
 
-// hasMorphism is
+// hasMorphism is the set of nodes that is reachable via either a *Path, a
+// single node.(string) or a list of nodes.([]string)
 func hasMorphism(via interface{}, nodes ...string) morphism {
 	return morphism{
 		Name:     "has",
--- a/graph/path/morphism_apply_functions.go
+++ b/graph/path/morphism_apply_functions.go
@@ -45,7 +45,7 @@ func isMorphism(nodes ...string) morphism {
 				return in
 			}
 
-			isNodes := ls.FixedIterator()
+			isNodes := qs.FixedIterator()
 			for _, n := range nodes {
 				isNodes.Add(qs.ValueOf(n))
 			}
@@ -78,7 +78,7 @@ func hasMorphism(via interface{}, nodes ...string) morphism {
 			}()
 
 			trail := iterator.NewLinksTo(qs, viaIter, quad.Predicate)
-			dest := iterator.NewLinksTo(qs, ends, graph.Object)
+			dest := iterator.NewLinksTo(qs, ends, quad.Object)
 
 			// If we were given nodes, intersecting with them first will
 			// be extremely cheap-- otherwise, it will be the most expensive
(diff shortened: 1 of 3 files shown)


kortschak:
Note that copyright dates are starting points of a time interval.

graph/path/morphism_apply_functions.go
```diff
@@ -20,58 +20,80 @@ import (
 	"github.com/google/cayley/quad"
 )
 
+// join puts two iterators together by intersecting their result sets with an AND
```
kortschak: "intersect" rather than "join"?

graph/path/morphism_apply_functions.go
```diff
@@ -20,58 +20,80 @@ import (
...
+			}
+
+			// Anything with fixedIterators will usually have a much
+			// smaller result set, so join isNodes first here
+			return join(qs, isNodes, in)
 		},
 	}
 }
 
+// hasMorphism is
```
kortschak: Complete this comment.

graph/path/morphism_apply_functions.go
Kavec: I was waffling between the two-- on one hand, we're logically intersecting two result sets here. On the other, the intention is to join two paths together. In that sense, I think calling this 'join' is a lot more evocative of intent, if not as clear on mechanics.

Kavec:
On copyrights, whoops! That's been fixed.

kortschak:
Would you mind terminating sentences with periods throughout?

graph/path/morphism_apply_functions.go
Kavec: Fixed. In my local copy this function got renamed to reaches, but I didn't want to break the API via renaming exported functions here. This one in particular is tricky to describe as Has-- thanks for the current review, and please let me know if I've misunderstood the logic this represents.

Kavec:
Done! I'll ensure further pull requests stick to that, too; what else needs to be touched up?

kortschak:
Looks generally OK to me.

barakmich:
LGTM, merging.

PR #315, opened 2015-10-06
Improve Transaction: deduplicate quads, allow adding/removing a quad in the same tx

Quentin-M:
It basically allows to add and remove the same quad in a transaction without failing and without depending on the underlying backend. See tests for more insights on what that's doing.

(CLA Accepted)

### DIFF
--- a/graph/transaction.go
+++ b/graph/transaction.go
@@ -16,19 +16,19 @@ package graph
 
 import "github.com/google/cayley/quad"
 
-// Transaction stores a bunch of Deltas to apply atomatically on the database
+// Transaction stores a bunch of Deltas to apply atomatically on the database.
 type Transaction struct {
 	Deltas map[Delta]struct{}
 }
 
-// NewTransaction initialize a new transaction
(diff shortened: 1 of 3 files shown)


graph/transaction.go
```diff
@@ -14,30 +14,52 @@
...
+	Deltas map[Delta]struct{}
 }
 
+// NewTransaction initialize a new transaction
 func NewTransaction() *Transaction {
-	return &Transaction{make([]Delta, 0, 5)}
+	return &Transaction{Deltas: make(map[Delta]struct{}, 100)}
 }
 
+// AddQuad adds a new quad to the transaction if it is not already present in it
```
jzelinskie: Add a period at the end of this sentence.

graph/transaction.go
```diff
@@ -14,30 +14,52 @@
...
+// If there is a 'remove' delta for that quad, it will remove that delta from
+// the transaction instead of actually addind the quad
```
jzelinskie: another period

barakmich:
LGTM

PR #316, opened 2015-10-06
sql: Combine AND(Fixed, SQL) into a single IN clause, reducing roundtrips.…

### LINKED ISSUE #314: Optimize SQL Save()
Multiple chained Save() should only generate one SQL request instead of one SQL query per Save()

An example of the current behavior can be found here : [http://x.quentin-machu.fr/odeniqavef.json](http://x.quentin-machu.fr/odeniqavef.json)

barakmich:
… Add a test and benchmark.

Benchmarks on localhost:
The new test before:
```
BenchmarkSaveBogartPerformances	     200	   5916752 ns/op
```

The new test after:
```
BenchmarkSaveBogartPerformances	     500	   3431612 ns/op
```

So about 2x. Expect a bigger win when talking to a remote server.

Fixes #314

### DIFF
--- a/graph/sql/optimizers.go
+++ b/graph/sql/optimizers.go
@@ -156,9 +156,12 @@ func (qs *QuadStore) optimizeLinksTo(it *iterator.LinksTo) (graph.Iterator, bool
 	switch primary.Type() {
 	case graph.Fixed:
 		size, _ := primary.Size()
+		if size == 0 {
+			return iterator.NewNull(), true
+		}
 		if size == 1 {
 			if !graph.Next(primary) {
-				panic("unexpected size during optimize")
+				panic("sql: unexpected size during optimize")
 			}
 			val := primary.Result()
 			newIt := qs.QuadIterator(it.Direction(), val)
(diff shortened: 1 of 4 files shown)


Quentin-M:
It does not optimize every cases yet : http://x.quentin-machu.fr/neroqavelu.json

Quentin-M:
LGTM.

Got 10x to 35x improvement from local <-> RDS ! Well done !

graph/sql/optimizers.go
```diff
@@ -169,6 +172,32 @@ func (qs *QuadStore) optimizeLinksTo(it *iterator.LinksTo) (graph.Iterator, bool
...
 			return newIt, true
+		} else {
+			if !graph.Next(primary) {
+				panic("unexpected size during optimize")
+			}
+			val := primary.Result()
+			newIt := qs.QuadIterator(it.Direction(), val)
+			linkit := newIt.(*SQLIterator).sql.(*SQLLinkIterator)
+			if len(linkit.constraints) != 1 {
+				panic("constraint list should be only one, after fresh creation")
```
jzelinskie: panics are prefaced with "packagename: " before the actual message

Quentin-M:
SHEEP IT

(LGTM)

PR #317, opened 2015-10-06
graph: Add transaction err for sql

barakmich:
Really, it'd be nice to have a multi-error for backends -- determining whether the backend connection is what's dead or if the ApplyDeltas call failed semantically (ie, which of the class ErrQuadExists)


kortschak:
Would you add a note to #157 about what you want here? I still have every intention of improving the error handling, it's just a matter of finding the time.

PR #320, opened 2015-10-19
Unique iterator doesn't work with BoltDB

mackross:
With BoltDB the graph.Value can be two different pointers with the same contents. This causes the unique iterator to fail.


dennwc:
Fixed with d42aa90c4c64542ff8bc982925c68bb802f9683f. But thanks anyway.

PR #323, opened 2015-10-24
Quad readers/writers and formats

dennwc:
### Idea

Implement simple `Reader` and `Writer` interfaces and helper functions like in `io` package, but for quads.

Code already contained interfaces like `graph.QuadWriter` and `quad.Unmarshaler`. First one is too wide to implement, and the second one is named strangely for a generic quad reader interface (imagine quad slice, that exposes `Unmarshal` function).

The second point is a support for new quad encoding and decoding formats. It will be easier to have a `quad.RegisterFormat` function, like in `image` package to allow anyone to register encoder and decoders of quad sets. It will also benefit from shared Reader and Writer interfaces.
Registration process can also expose additional metadata about formats: file extensions and MIME types. This allows to pick format automatically using file path, or serve/load quads with a right format using HTTP Accept/Content-Type headers.
#### Core
- Added `quad.Writer` and `quad.BatchWriter` interfaces to replace `AddQuad` and `AddQuadSet` from `graph.QuadWriter`. Code updated to use new functions.
- Added `quad.Reader` interface to replace `quad.Unmarshaler`. Code updated.
- Added `Copy` and `CopyBatch` functions to simplify transfer between Reader and Writer. Loader uses these functions now. `Copy` will try to type-assert source and destination to Batch interfaces to speedup transfer process. 
- Added few functions to convert `graph.QuadStore`, `graph.Iterator` and `[]Quad` to `quad.Reader`.
- Added `quad.RegisterFormat` to make it easy to add new encoding/decoding formats.
  All export functions moved from exporter.go to sub packages in quad package and registered as encoding formats. All encoders/decoders implements Reader or Writer interface.
#### Formats
- Rewrote GML and GraphML encoder to write files in one pass.
- Added JSON-LD encoding/decoding support.
#### Cayley binary
- Select load/dump quad serialization format from registered ones.
- Determine quad format using Content-Type header in WriteQuad HTTP API.

(I've signed CLA)

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/323)


kortschak:
I think this requires more discussion. On the face of it, I don't like it much, but there is too much going on at a variety of different levels to be able to discuss it cleanly. There are clearly useful things here, but I think it really should be broken up into a number of PRs or at the very least, commits.

I'm also troubled by the number of errorable calls that ignore returns. This happens elsewhere, but the count of these should be reduced, not increased.

dennwc:
@kortschak In fact, this is not a huge change. There is a few old functions with a new names (old ones are preserved to guaranty compatibility), and a more standard way to deal with export/import functionality. Anyway, I can comment on these changes in more details, if you'd like.

And yes, I've also noticed a lot of `AddQuad` calls, mostly from tests, that ignores returned errors. Did you meant that I also should take care of them?

kortschak:
There are a few things going on. Some are name changes which may or may not be needed (and break backwards compatibility) and there are some large behavioural changes like registration - which I don't like* - I think these should not be happening together.

I don't think you need to fix up missed error checking for things outside the change set, but if you are changing code it should be improved at the same time.

\*An explanation of why I don't like the registration is that it essentially ties the quad parsers to cayley where at the moment they are entirely usable as standalone packages. It seems to me that this kind of thing should be done at the application layer, not the library layer. My interest in cayley is not via the command but via the back end, this kind of change complicates my life and potentially results in additional imports that I may not want in the application I am building.

dennwc:
This makes sense. But I've also assumed, that registration should not bring any new import dependencies. For example, if you like to import nquads from somewhere, the parser that you are using need to know what output format it should provide. For Cayley it's definitely a set of quad.Quad objects. So you'll have indirect dependency on quad package in any case. The only change I've made additionally, is that if you import nquads package, you can also get some metadata for that format (file extension, MIME type). It will not use any other imports, but will provide a helpers for user code to select one of registered formats using name/mime/file extension. But it will help to load and dump quad files, and send them via HTTP with a correct Content-Type header if user decide to implement such functionality on his own.

As the second layer, we could add a package to open file from FS using only path and let it guess a quad encoding format itself. Or we may add an custom HTTP Get function targeted for Semantic Web, which adds a proper Accept header using known quad encoding formats. But this may be undesirable functionality, as you pointed out. And there is no such functions in the PR.

So, I believe it is a pretty safe change from that perspective.

But the question with function names stays open. I leaved old interface functions untouched, but if you had a custom implementation of `QuadWriter`, you'll need to implement a one line wrapper function for new interface. The second change is that `Load` and `DecompressAndLoad` now accepts a new interface as the parameter for callback, so if you had a custom loading logic, you'll have to rename that callback parameter as well.

dennwc:
Merged upstream changes. Is there any updates on this?

dennwc:
@kortschak Would you mind looking through the new change set? I've integrated format selectors to CLI and to HTTP API. Now both write and delete endpoints of API can select the right quads serialization format depending on a Content-Type header (or fallback to default JSON format). Would it be helpful to implement a similar API and behavior for dumping the database?
Also, please take a look at the `quad.Copy` function. It definitely makes a life a lot easier, when using Cayley as a library. It can automatically cast src/dst to a Batch versions of the interface and speedup an import/export process, for example.

@barakmich What do you think about renaming of `AddQuad` and `AddQuadSet` functions?

Btw, this PR addresses https://github.com/google/cayley/issues/176 a bit more. Proposed design can be extended a lot easier.

There was also a question in https://github.com/google/cayley/pull/292 regarding "loading a graph to dump it (in a different format)", and the same `quad.Copy` function can be used there to omit loading the graph. It will convert formats on the fly.

I'm also interested in integration of JSON-LD format. What is the preferred approach: either to bring an optional dependency on a lib like [goLD](https://github.com/kazarena/json-gold) or to embed a necessary code into Cayley source tree?

kortschak:
I can take a look at this next week.

graph/quadwriter.go
```diff
@@ -69,11 +69,18 @@ var (
 )
 
 type QuadWriter interface {
+	quad.Writer
+	quad.BatchWriter
+
 	// Add a quad to the store.
+	//
+	// Deprecated: use WriteQuad instead.
```
kortschak: If this name change is going to happen. Can we just delete the old methods. The longer something lies around, the greater the chance someone will use the deprecated method. Except where bound by backwards compatibility guarantees, deprecation is not a normal Go idiom.

\cc @barakmich

internal/dump.go
```diff
@@ -1,13 +1,18 @@
...
 	"fmt"
+	"io"
 	"os"
-	"compress/gzip"
 	"path/filepath"
 
 	"github.com/google/cayley/graph"
-	"github.com/google/cayley/exporter"
+	"github.com/google/cayley/quad"
+	_ "github.com/google/cayley/quad/gml"
```
kortschak: Add blank line and comment explaining '_' imports.

quad/cquads/cquads.go
```diff
@@ -45,8 +54,8 @@ func NewDecoder(r io.Reader) *Decoder {
 	return &Decoder{r: bufio.NewReader(r)}
 }
 
-// Unmarshal returns the next valid N-Quad as a quad.Quad, or an error.
-func (dec *Decoder) Unmarshal() (quad.Quad, error) {
```
kortschak: I'm still not clear why this change is required.

quad/jsonarr/jsonarr.go
```diff
@@ -0,0 +1,96 @@
+package jsonarr
```
kortschak: Just call this json. There's no collision within the cayley tree AFAICS and others can rename.

internal/load.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/google/cayley/internal/db"
 	"github.com/google/cayley/quad"
 	"github.com/google/cayley/quad/cquads"
+	_ "github.com/google/cayley/quad/jsonarr"
```
kortschak: Drop this to the bottom of the import block separated and comment.

quad/formats.go
```diff
@@ -0,0 +1,64 @@
+package quad
```
kortschak: Package comment.

quad/gml/gml.go
```diff
@@ -0,0 +1,98 @@
+package gml
```
kortschak: Package comment.

quad/graphml/graphml.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if w.err = xml.EscapeText(w.w, []byte(s)); w.err != nil {
+		return -1
+	}
+	if _, w.err = w.w.Write([]byte("</data></node>\n")); w.err != nil {
+		return -1
+	}
+	return i
+}
+func (w *Writer) WriteQuad(q quad.Quad) error {
```
kortschak: New line.

quad/cquads/cquads.go
dennwc: To clarify the naming. There is a sliceReader type at rw.go:51, for example, that is used in tests to load small quad sets. It will be weird to have an Unmarshal method on a slice. There is also a comment in dump.go about dumping the results of specific query only, so we might have a ReadQuad implementation that wraps the iterator too. And again, Unmarshal is a strange method for iterator.

quad/formats.go
dennwc: Small comment is already in place in quad.go. Are you asking about extending it?

graph/quadwriter.go
dennwc: It would be great to remove the old ones, but I think that if someone have used Cayley as a library, he definitely used Add methods as they were the fastest way to load quads. It will break code for sure. But from change perspective, renaming is far less painful comparing to signature change.

barakmich:
I'd agree with @kortschak -- deprecating/breaking current for a cosmetic change (AddQuad/AddQuadSet/Unmarshal to their new equivelents) isn't worth the benefit -- if there were a larger reason (implement a standard library interface or similar) then I might see an argument

Overall, adding a bunch of loaders/exporters is great -- `exporters/` is no problem at all. 

I'd be curious to re-review pending reverting the renamed functions

dennwc:
@barakmich Is there a chance to preserve both names? I mean, that current user's code could potentially use exported things: nquads decoder (renamed Unmarshal method) and "single" quad writer (AddQuad methods). This is only to structs, so we can theoretically keep old methods, and proxy them to a new ones (as this PR does) to preserve compatibility. All new formats and writer implementations can use new names without any impact. At least I believe that Unmarshaler interface is not used by users at all, because there was no format registration and all calls to quad decoder/encoder directly uses concrete struct types.

barakmich:
Sure; but if we have the functions already, and they have the same signature, why proxy? Then throughout our codebase we have two names for the same thing, which would probably best be avoided.

dennwc:
Yes, but proxy code can be told to soon-to-be-deprecated, and all new implementations can omit these, slowly decreasing the number of usages.

So, to clarify:
1) I can preserve ReadQuad name, because nobody uses Unmarshaler, right?
2) Without considering the breakage and transition, what names are clearer? Just for statistics)
3) Having two methods in core interface is bad. But for the time of early releases, we can make API better with the good naming. With later releases it would become impossible. Users will slowly switch from direct add-each-quad-loop (old AddQuad) method to a new Copy (which is smart enough to use Batch automatically). After that direct usage of AddQuad can become extremely rare, so we can deprecate it after some time. So the question: are we willing to make it now? Or (if answer 2 is positive) it will wait for a next major release that can break things? :)

And sorry for the buzz - I'm just trying to preserve some semantic, that is familiar to everyone in Go world :)

barakmich:
Reviewed 62 of 62 files at r1.

---

*[graph/quadwriter.go, line 236 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-KVqwXXwPGsANgUNyYiM:-KVqwXXwPGsANgUNyYiN:b1rwzlu) ([raw file](https://github.com/cayleygraph/cayley/blob/71252b3df233d4b96535894f6db2cab201bf2f20/graph/quadwriter.go#L236)):*
> ```Go
> 
> func (r *quadReader) ReadQuad() (quad.Quad, error) {
> 	if r.it.Next() {
> ```

This works when it's an AllIterator, or a top level quad iterator, but the (somewhat more useful) general case is to pull out all the quads traversed in a query. (Deduplication can be another reader above). Not saying this is wrong, however -- so I'd suggest the following (Names only a suggestion)

NewQuadStoreReader(qs QuadStore) -- equivalent to NewReader(qs, nil)
NewResultReader(qs QuadStore, it Iterator) -- equivalent to NewReader above
NewSubgraphReader(qs QuadStore, it Iterator) -- tags all iterators that point to quads in the query, and streams them out for every Next() and NextResult()
NewReader(qs QuadStore, it Iterator) -- the  above *and* dedupes them

The latter two can be a followup issue; but appropriate naming for what exists is what this comment is for.

---

*[quad/pquads/quads.proto, line 29 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-KVqyB0hiFXRwh6UC2zs:-KVqyB0hiFXRwh6UC2zt:blmel5d) ([raw file](https://github.com/cayleygraph/cayley/blob/71252b3df233d4b96535894f6db2cab201bf2f20/quad/pquads/quads.proto#L29)):*
> ```protobuf
>   string object = 3;
>   string label = 4;
>   Value subject_value = 5;
> ```

Observation: if we're going to extend the Quad message (and we should!) then subject_value and predicate_value isn't quite a value -- it's a bnode or IRI. If we further want to extend this (as we might, if we want a superset of RDF while thinking on reification) then we need either (a) another proto altogether or (b) can extend this one -- but for the moment it's safe (and probably wise) to restrict the set as per the spec

Value is correctly defined for object_value :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/323)*

dennwc:
Review status: 45 of 56 files reviewed at latest revision, 2 unresolved discussions.

---

*[graph/quadwriter.go, line 236 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-KVqwXXwPGsANgUNyYiM:-KVsk-ICpxsJ-AkxkYOr:b-r3gouv) ([raw file](https://github.com/cayleygraph/cayley/blob/71252b3df233d4b96535894f6db2cab201bf2f20/graph/quadwriter.go#L236)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> This works when it's an AllIterator, or a top level quad iterator, but the (somewhat more useful) general case is to pull out all the quads traversed in a query. (Deduplication can be another reader above). Not saying this is wrong, however -- so I'd suggest the following (Names only a suggestion)
> 
> NewQuadStoreReader(qs QuadStore) -- equivalent to NewReader(qs, nil)
> NewResultReader(qs QuadStore, it Iterator) -- equivalent to NewReader above
> NewSubgraphReader(qs QuadStore, it Iterator) -- tags all iterators that point to quads in the query, and streams them out for every Next() and NextResult()
> NewReader(qs QuadStore, it Iterator) -- the  above *and* dedupes them
> 
> The latter two can be a followup issue; but appropriate naming for what exists is what this comment is for.
</details>
Done.

But I don't see how latter two should work. If we walk an iterator tree we can find a links iterator at some level, but how to determine if it's the right one? Most of them are used with conjunction with And, so we should add a heuristic to tag And (and some others?) instead of LinksTo. Or should we check if iterator was tagged by user before adding it as Read source? I agree that it should be a separate issue.

---

*[quad/pquads/quads.proto, line 29 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-KVqyB0hiFXRwh6UC2zs:-KVsmI34dRLLwetE1wd3:bao8tbg) ([raw file](https://github.com/cayleygraph/cayley/blob/71252b3df233d4b96535894f6db2cab201bf2f20/quad/pquads/quads.proto#L29)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> Observation: if we're going to extend the Quad message (and we should!) then subject_value and predicate_value isn't quite a value -- it's a bnode or IRI. If we further want to extend this (as we might, if we want a superset of RDF while thinking on reification) then we need either (a) another proto altogether or (b) can extend this one -- but for the moment it's safe (and probably wise) to restrict the set as per the spec
> 
> Value is correctly defined for object_value :)
</details>
Agreed, but Cayley accepts any values for each direction, so this format write them as-is. We should disallow string literals in subject for all parsers, then.

To make it a separate issue, I've added an StrictQuad definition which is written according to RDF and a WireQuad which is current Cayley format (if we drop legacy fields from proto). Implementation has an option to write in strict mode, and there is a separate flag in file header to indicate that. If we decide to go RDF-only, we can drop this flag and definition of WireQuad and change nothing in format itself.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/323)*

barakmich:
Reviewed 1 of 62 files at r1, 11 of 11 files at r2.
Review status: 55 of 56 files reviewed at latest revision, 1 unresolved discussion.

---

*[quad/rw.go, line 45 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-KWUkQ8PSAMg8RYwSpjA:-KWUkQ8PSAMg8RYwSpjB:bg7bbpz) ([raw file](https://github.com/cayleygraph/cayley/blob/529bcee69633944265e1d58b720e257f46c5fee9/quad/rw.go#L45)):*
> ```Go
> type ReadSkipCloser interface {
> 	Reader
> 	Skiper
> ```

Interesting semantic thing; Go says it's always `Xer` But `Skipper` is the spelling. I'd go with `xSkipper`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/323)*

barakmich:
:lgtm:

---

Review status: 55 of 56 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/323#-:-KWUmTaIzPpRrRSzrEU_:bnfp4nl)*

PR #330, opened 2015-10-29
query/gremlin: Refactor in terms of graph/path

barakmich:
We were building iterators the same way in two places; the gremlin-esque path library and the Gremlin JS library. The latter now depends on the former.

This means two really great things: 1) the path library now has feature parity with the JS (yay) and 2) new features only need a single implementation in Go, and can easily be exposed in javascript.

The new context being routed through path.Path is in pursuit of a future feature; I started trying to do that and got diverted into doing a refactor.

### DIFF
--- a/graph/path/path.go
+++ b/graph/path/path.go
@@ -20,13 +20,35 @@ type applyMorphism func(graph.QuadStore, graph.Iterator, *context) (graph.Iterat
 
 type morphism struct {
 	Name     string
-	Reversal func() morphism
+	Reversal func(*context) (morphism, *context)
 	Apply    applyMorphism
 	tags     []string
+	context  context
 }
 
+// context allows a high-level change to the way paths are constructed. Some
+// functions may change the context, causing following chained calls to act
+// cdifferently.
+//
+// In a sense, this is a global state which can be changed as the path
+// continues. And as with dealing with any global state, care should be taken:
+//
+// When modifying the context in Apply(), please copy the passed struct,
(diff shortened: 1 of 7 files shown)


graph/path/path.go
```diff
@@ -16,18 +16,25 @@ package path
...
+
 type morphism struct {
 	Name     string
 	Reversal func() morphism
-	Apply    graph.ApplyMorphism
+	Apply    applyMorphism
 	tags     []string
 }
 
+type context struct {
```
kortschak: Probably worth adding some documentation on the intention of this type.

graph/path/path.go
```diff
@@ -111,7 +118,61 @@ func (p *Path) Out(via ...interface{}) *Path {
...
+// OutWithTags is exactly like In, except it tags the value of the predicate
+// traversed with the tags provided.
+func (p *Path) OutWithTags(tags []string, via ...interface{}) *Path {
+	p.stack = append(p.stack, outMorphism(tags, via...))
+	return p
+}
+
+// Both updates this path following both inbound and outbound predicates.
+//
+// For example:
```
kortschak: For the code examples here and below, add an additional space before the code snippet on each line to get a `<pre>` block.

kortschak:
LGTM with minor comments.

kortschak:
Very nice.

barakmich:
Fixed. Merging! (Next up, using that context)

PR #336, opened 2015-11-16
Use a Size() estimate for SQL Quadstore when noSizes is enabled.

Quentin-M:
Counting on PostgreSQL could be really slow because of its MVCC model,
if noSizes is enabled, we should use an estimate

Ref: https://wiki.postgresql.org/wiki/Count_estimate

(CLA Accepted)

### DIFF
--- a/graph/sql/quadstore.go
+++ b/graph/sql/quadstore.go
@@ -37,11 +37,12 @@ var (
 )
 
 type QuadStore struct {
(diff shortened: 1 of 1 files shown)


barakmich:
There we go. If you can file an issue to merge noSizes and useEstimates and maybe figure out a faster more accurate way to do things, that'd be a big help. Otherwise, merging.

PR #337, opened 2015-11-17
Ensure that quads in a transaction are applied in the desired order.

Quentin-M:


graph/transaction.go
```diff
@@ -52,14 +48,41 @@ func (t *Transaction) AddQuad(q quad.Quad) {
...
+	if _, adExists := t.deltas[ad]; adExists {
+		t.deleteDelta(ad)
+	} else {
+		if _, rdExists := t.deltas[rd]; !rdExists {
+			t.addDelta(Delta{Quad: q, Action: Delete})
+		}
+	}
+}
+
+func createDeltas(q quad.Quad) (ad, rd Delta) {
```
barakmich: There's no need to make two. Just pass it by value, and it will copy

graph/transaction.go
```diff
@@ -52,14 +48,41 @@ func (t *Transaction) AddQuad(q quad.Quad) {
...
+		Action: Delete,
+	}
+	return
+}
 
-	if _, adExists := t.Deltas[ad]; adExists {
-		delete(t.Deltas, ad)
-	} else {
-		t.Deltas[Delta{Quad: q, Action: Delete}] = struct{}{}
+func (t *Transaction) addDelta(d Delta) {
```
barakmich: (copy, eg, here. this you do need)

graph/transaction.go
```diff
@@ -52,14 +48,41 @@ func (t *Transaction) AddQuad(q quad.Quad) {
...
+	t.Deltas = append(t.Deltas, d)
+	t.deltas[d] = struct{}{}
+}
+
+func (t *Transaction) deleteDelta(d Delta) {
```
barakmich: One idea would be to make the map a `map[delta]int` -- and keep the index.

barakmich: But that's a bad idea because updating everything would be a pain. As-is is good.

barakmich:
LGTM on green

PR #324, opened 2015-10-25
Update docs for PostgreSQL configuration

duncanawoods:

PR #341, opened 2015-11-18
Docs

marcusmonteiro:
I wrote an `init` command example for the Postgres backend in the Overview doc, and a link for the Postgres site in README.

I didn't edit the Configuration doc because the `database` section contains technical information that I don't know about, I didn't know what to write in the `db_path` section, and I don't know if there are any `Per-Database Options` for Postgres databases.

I signed the CLA.

### DIFF
--- a/docs/Overview.md
+++ b/docs/Overview.md
@@ -19,6 +19,7 @@ Examples for each backend:
   * `leveldb`:  `./cayley init --db=leveldb --dbpath=/tmp/moviedb` -- where /tmp/moviedb is the path you'd like to store your data.
   * `bolt`:  `./cayley init --db=bolt --dbpath=/tmp/moviedb` -- where /tmp/moviedb is the filename where you'd like to store your data.
   * `mongo`: `./cayley init --db=mongo --dbpath="<HOSTNAME>:<PORT>"` -- where HOSTNAME and PORT point to your Mongo instance.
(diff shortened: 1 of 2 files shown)


docs/Overview.md
```diff
@@ -19,6 +19,7 @@ Examples for each backend:
   * `leveldb`:  `./cayley init --db=leveldb --dbpath=/tmp/moviedb` -- where /tmp/moviedb is the path you'd like to store your data.
   * `bolt`:  `./cayley init --db=bolt --dbpath=/tmp/moviedb` -- where /tmp/moviedb is the filename where you'd like to store your data.
   * `mongo`: `./cayley init --db=mongo --dbpath="<HOSTNAME>:<PORT>"` -- where HOSTNAME and PORT point to your Mongo instance.
+  * `sql`: `./cayley init --db=sql --dbpath="user=user dbname=mydb password=mypassword sslmode=disable"` -- where the value of the `dbpath` flag is a connection string of parameters from [pq](https://github.com/lib/pq). See https://godoc.org/github.com/lib/pq for more information.
```
mwmahlberg: Is `sslmode=disable` necessary, strictly speaking? Have you tried it with ssl enabled?

mwmahlberg:
In which way is your PR different from #324 ?

marcusmonteiro:
I don't remember if it was necessary. It solved the problem I was having at
http://stackoverflow.com/questions/33769116/how-to-use-cayley-with-a-postgres-backend.
Back then `init` didn't create the necessary tables, I don't know if they
changed something at cayley or libpq since then. As an opinion, I think
that referencing the libpq godocs is nice, because it took me a while to
discover libpq starting from the cayley docs.

robertmeta:
@marcusmonteiro Thanks so much for the PR, we added it to the other one (specifically the important bit about where dbpath comes from).

PR #345, opened 2015-11-25
add Triple function

jzelinskie:
This is a shortcut for those that are not using the label field. It can
become very verbose always calling Quad with an empty string as the last
parameter.

I've signed the CLA.


barakmich:
LGTM

PR #347, opened 2015-11-25
initial Dockerfile

jzelinskie:


Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:latest
+MAINTAINER Barak Michener <me@barakmich.com>
+
+# Set up workdir
+WORKDIR      /go/src/github.com/google/cayley
+RUN mkdir -p /go/src/github.com/google/cayley
```
tianon: `WORKDIR` will auto-create, so this line is unnecessary :smile:

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Restore vendored dependencies
+RUN go get github.com/tools/godep
+ADD Godeps /go/src/github.com/google/cayley/Godeps
+RUN godep restore
+
+# Add and install cayley
+ADD . .
+RUN go install github.com/google/cayley/cmd/cayley
```
tianon: Any reason not to just `./...` here to make sure you capture any future additions without changes?  Also, I'm personally a fan of including `-v` here so you know what it's doing while it sits and compiles, but YMMV. :+1:

imports.go
```diff
@@ -27,6 +27,10 @@ type Handle struct {
 	graph.QuadWriter
 }
 
+func Triple(subject, predicate, object string) quad.Quad {
+	return quad.Quad{subject, predicate, object, ""}
+}
+
```
paultag: Looks like this landed on the wrong branch

Dockerfile
jzelinskie: I've added -v, but you cannot use "./..." without getting a failure because the `integration` directory contains no buildable source files.

tianon: Ah, ok, so maybe just "github.com/google/cayley/cmd/..." ?

barakmich:
LGTM

PR #348, opened 2015-11-25
graph: make quadstore init functions idempotent

barakmich:
For use in #347, as well as general user-friendliness and cleanup.


jzelinskie:
LGTM

PR #368, opened 2016-01-23
path: Concrete types (for optimization and serialization)

dennwc:
## Preface

_(I may update header from time to time to reflect changes in the code)_

For now path lib is only used as a wrapper around iterators, but I think it deserves a separate handling mechanism. This PR proposes a concrete types for path elements (morphisms). Some pros of the change:
- Most of the time we don't need to optimize iterators shape - we can optimize paths shape instead and then spawn already-optimized iterators from them.
- Paths can be reused. Thus, user can optimize most-used query once and use it faster.
- Paths are state-less. We can define a serialization mechanism (protobuf?) which will allow to construct queries separately (client-side), and then pass resulting path to the server. This can lead to more extensible query layer.
- As an additional optimization, we can "bind" path to a specific quad store. For example, SQL implementation may replace some path elements with a plain-text SQL queries (or at least fast iterator generators). Each iterator, produced from such path will be already optimized for a given quad store.
- Keeping in mind two previous statements, we can define a function to convert iterators back to paths, so they can be serialized again. Then we can pass serialized queries back to client as pagination tokens (each iterator implementation will insert "Skip" path element if it was already iterated to some extent).
- Concrete types can be processed with reflection allowing operations like "find all Placeholder paths in this tree and replace them with ...".
- Optimization code is easier and allows to use high-level constructs.
## Implementation

For now, code is located in a separate folder `graph/path2` for evaluation purposes. It preserve compatibility with current path lib implementation, but misses a "Back" feature (will be implemented soon). It may also miss some optimizations.
### Path interfaces

Path interface is pretty generic:

``` Go
type PathObj interface{
    BuildIterator() graph.Iterator
}
```

For type-safety reasons it was split into two sub-interfaces:

``` Go
type Nodes interface {
    PathObj
    Optimize() (Nodes, bool)
    // ...
}
type Links interface {
    PathObj
    Optimize() (Links, bool)
    // ...
}
```

With this approach, path types cannot be mixed, making it a compilation error. One may complain about code duplication, but most of existing iterators falls into Nodes category, so the cost of such decision is minimal.
### Abstract paths

Some path elements are "abstract" - they cannot generate iterators without a concrete quad store (AllNodes, AllLinks for example). There is a optional interface for such types:

``` Go
type NodesAbstract interface {
    BindTo(qs graph.QuadStore) Nodes
}
type LinksAbstract interface {
    BindTo(qs graph.QuadStore) Links
}
```

Those path elements should be replaced with a concrete variant by quad store or by replacing function. Without it, they may panic on iterator generation.

**Current list of abstract paths:**
- AllNodes
- AllLinks
- NotNodes (for optimization purposes)
- Fixed (should be converted into graph.Value by QS)
- HasA (not abstract, but constructor requires QS to call Contains)
- LinksTo (same reason)
### Basic paths

These are converted from original Cayley iterators, thus they have implementation independent from a particular quad store. Most of them are one-liners: call iterator constructor. Some examples:

``` Go
type Tag struct{
    Nodes Nodes
    Tags []string
}

type IntersectNodes []Nodes

type Unique struct {
    Nodes Nodes
}
```
### High-level paths

These are converted from current path lib. The idea here is to preserve a high-level view on path to allow easier optimization code, but still having a low-level fallback variant. Such paths can implement an optional interface:

``` Go
type NodesSimplifier interface {
    Simplify() Nodes
}
// Same duplicated for Links
```

"Simplify" function tells path element to build an analog of itself in terms of basic path elements. This mechanism is also used as a second pass of optimization: firstly let high-level iterator to apply some optimization and if it fails, simplify path to a tree of basic paths and try to optimize it again. Full optimization pass is described below.

**High-level path example:**

``` Go
type Predicates struct {
    From Nodes
    Rev   bool
}
func (p Predicates) Simplify() Nodes {
    dir := quad.Subject
    if p.Rev {
        dir = quad.Object
    }
    return Unique{
        Nodes: HasA{
            Links: LinksTo{
                Nodes: p.From,
                Dir:   dir,
            },
            Dir: quad.Predicate,
        },
    }
}
```
### Replacers

An ability to replace some path elements in tree is important for easy manipulation, thus a Replace method is required for paths:

``` Go
type NodesWrapper func(Nodes) Nodes
type Nodes interface {
    // ...
    Replace(NodesWrapper, LinksWrapper) Nodes
}
// The same is duplicated for Links
```

Method tells a path to generate a clone of itself, applying wrapper functions for sub-paths with appropriate types.
**TBD:** We can make a generic reflect-based replacer instead of forcing implementation of Replace function for every path type.

This replacement mechanism allows to replace Abstract paths with a concrete implementations, allows to use pure morphisms (paths without a start, that can be followed from any other path point), simplifies high-level path elements that was not optimized by QS and much more.
### Optimization steps

**1) Pure path optimization**
This step calls Optimize method for each path element. Paths can be moved around, merged or disappear, but abstract and high-level paths will preserve their own type. Example: optimize ordering and merge "Fixed" nodes in "Intersect" ("And") path element.

**2) QS optimization**
Step lets quad store to optimize path. It will try to optimize high-level paths first and if it fails, path will be simplified, optimized according to step 1 and QS will retry optimization. Example: QS may optimize "Out" path at once, or simplify it and optimize all "LinksTo" in there.

**3) Bind path to QS**
The last step, which involves replacing abstract path elements with iterator constructor for a given QS, allowing to call BuildIterator for the whole tree.
### End notes

Any feedback is welcomed. I will greatly appreciate any help on concrete path types for SQL implementation and path optimizations in general.

More tests and benchmarks will appear soon.
As a raw result: my benchmark showed 20% speedup on memstore in "the helpless checker" integration test. But it may be related to some other cause, which I might be not aware of.

PR also includes few unrelated things, which were included to fix a specific bug or for convenience:
- Fix for Unique iterator for LevelDB/Bolt (https://github.com/google/cayley/commit/0158b23da11f688c85a8ed796ea67c5b61847019, the same as https://github.com/google/cayley/pull/320)
- Fix Postgres enumerating empty quad label in AllNodes iterator (https://github.com/google/cayley/commit/c1ff6ac7a866c27103c021ef3e61e42a62871532)
- Some code to bring up a fresh Mongo/Postgres in Docker for tests (will be skipped if it's not installed) (https://github.com/google/cayley/commit/b2523d17fa28b194a420365687f5e4f5db77ccd1 and https://github.com/google/cayley/commit/a2a3e136384a5039481e767239dcd680d38572fe)
#### TODOs:
- [ ] Concrete path types for SQL backend.
- [ ] Docs on path types.
- [ ] More tests.
- [ ] Benchmarks.
- [ ] Define a cost mechanism for paths (if possible).
- [ ] Port other optimizations.
- [ ] Implement Back mechanism for compatibility layer.
- [ ] Resolve reflect-vs-interface question for Replacer.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -68,6 +68,10 @@
 			"ImportPath": "gopkg.in/mgo.v2",
 			"Comment": "r2015.01.24",
 			"Rev": "c6a7dce14133ccac2dcac3793f1d6e2ef048503a"
+		},
+		{
+			"ImportPath": "github.com/fsouza/go-dockerclient",
+			"Rev": "25bc220b299845ae5489fd19bf89c5278864b050"
 		}
 	]
 }
--- a/graph/bolt/bolt_test.go
+++ b/graph/bolt/bolt_test.go
@@ -24,9 +24,36 @@ import (
 
 	"github.com/google/cayley/graph"
 	"github.com/google/cayley/graph/iterator"
+	"github.com/google/cayley/graph/path2/pathtest"
 	"github.com/google/cayley/quad"
 	"github.com/google/cayley/writer"
 )
+
+func makeStore(t testing.TB) (graph.QuadStore, func()) {
+	tmpFile, err := ioutil.TempFile(os.TempDir(), "cayley_test")
+	if err != nil {
+		t.Fatalf("Could not create working directory: %v", err)
+	}
+	t.Log(tmpFile.Name())
+
+	closer := func() {
+		os.RemoveAll(tmpFile.Name())
+	}
+
+	err = createNewBolt(tmpFile.Name(), nil)
+	if err != nil {
+		closer()
+		t.Fatalf("Failed to create Bolt database.")
+	}
+
+	qs, err := newQuadStore(tmpFile.Name(), nil)
+	if qs == nil || err != nil {
+		closer()
+		t.Fatal("Failed to create Bolt QuadStore.")
+	}
+
+	return qs, closer
+}
 
 func makeQuadSet() []quad.Quad {
 	quadSet := []quad.Quad{
@@ -503,4 +530,8 @@ func TestDeletedFromIterator(t *testing.T) {
 	if got := iteratedQuads(qs, it); !reflect.DeepEqual(got, expect) {
 		t.Errorf("Failed to get expected results, got:%v expect:%v", got, expect)
 	}
+}
+
+func TestMorphisms(t *testing.T) {
+	pathtest.TestMorphisms(t, makeStore)
 }
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/google/cayley/graph"
 	"github.com/google/cayley/graph/iterator"
+	"github.com/google/cayley/graph/path2"
 	"github.com/google/cayley/graph/proto"
 	"github.com/google/cayley/quad"
 )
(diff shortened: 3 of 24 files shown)


dennwc:
Will rewrite this part after reification. Closing.

PR #374, opened 2016-02-19
Added basic RPM packaging

mwmahlberg:
Added basic RPM packaging, which works only on tags, currently, so the RPM is only easily built when it is included within the tag to build.
# How to use

Your RPM build tree should as always look like this

```
./
├── BUILD
├── BUILDROOT
├── RPMS
├── SOURCES
├── SPECS
└── SRPMS
```

From there, the package is easy to build

```
cd SOURCES
curl -L https://github.com/google/cayley/archive/v0.4.2.tar.gz -o cayley-0.4.2.tar.gz
rpmbuild -ta cayley-0.4.2.tar.gz
```
# Release

The version number in `_rpm/cayley.spec` needs to be bumped manually. I will try to find a way to automate this. Meanwhile, the line

```
Version: 0.4.1
```

needs to be adjusted on release.
# Documentation

I will write a more detailed documentation in the wiki
# CLA

The CLA is signed as mwmahlberg.

### DIFF
--- a/_rpm/cayley.env.conf
+++ b/_rpm/cayley.env.conf
@@ -1,0 +1,9 @@
+[Service]
+Environment=MODE=http
+Environment=PORT=64210
+Environment=HOST=0.0.0.0
+Environment=DB=bolt
+Environment=DBPATH=/var/lib/cayley/cayley.db
+
+# Uncomment and add additional flags here, if necessary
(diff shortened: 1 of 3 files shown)


dennwc:
Closing in favor of #455

Issue #290, opened 2015-07-31
Expose Value Comparison Iterator in Gremlin


barakmich:
We already have the iterator, it supports strings as of #262, I'm open for proposals to add it to Gremlin; it shouldn't be too difficult.

ds--:
We don't have anything along Gremlin's normal [Has](http://gremlindocs.com/#filter/has)/[HasNot](http://gremlindocs.com/#filter/hasnot).

ds--:
I have a little followup,

I've recently been experimenting (2fd2be51f07b349b09157aab452e18d9bb08c4ed) with that comparison iterator and extending Has() so that it maintains its existing behavior for backwards compatibility but now allows arrays for predicate/object, example:
* `Who follows dani or greg?` can be more concisely expressed as `g.V().Has("follows", ["dani", "greg"])`.
* `Who follows or likes dani or greg?` can be expressed as `g.V().Has(["follows", "likes"], ["dani", "greg"])`

Additionally I've expended it to also support the [full Gremlin syntax](http://gremlindocs.com/#filter/has) for `Has()` when 3 arguments are provided `Has(<predicate>, <operator>, <value[s]>)` For example:
* `Who is over 25?`, assuming our test data had a few age predicates, would be expressed as `g.V().Has("age", T.gt, 25)`
* `Who is 18 or 21?` can be expressed as `g.V().Has("age", T.in, [18, 21])`

With a few additions to allow a few more text related comparisons: (for example case insensitivity here)
* `Who is from the US or UK?` would be `g.V().Has("country", T.Text.iin, ["UK", "USA"])`

I am looking for some feedback on everything (syntax, code itself). Looking for some approval on this direction.

PR #378, opened 2016-02-28
Typed Quad values

dennwc:
### Strictly-typed quad values

The idea is to change string type of quad directions to interface-based Value, to allow to preserve types of subject and object particularly.

```Go
type Value interface {
	String() string
}

type Quad struct {
	Subject   Value
	Predicate Value
	Object    Value
	Label     Value
}
```

As a part of effort, unify tests for QuadStore to check that all implementations behaves in a similar way.

### Supported types:
```Go
// Raw is a Turtle/NQuads-encoded value.
type Raw string

// String is an RDF string value (ex: "name").
type String string

// TypedString is an RDF value with type (ex: "name"^^<type>).
type TypedString struct {
	Value String
	Type  IRI
}

// LangString is an RDF string with language (ex: "name"@lang).
type LangString struct {
	Value String
	Lang  string
}

// IRI is an RDF Internationalized Resource Identifier (ex: <name>).
type IRI string

// BNode is an RDF Blank Node (ex: _:name).
type BNode string

// Int is a native wrapper for int64 type.
type Int int64

// Float is a native wrapper for float64 type.
type Float float64

// Bool is a native wrapper for bool type.
type Bool bool

// Time is a native wrapper for time.Time type.
type Time time.Time
```
### TODO:
- [x] Value definition
- [x] Update cquads and nquads parsers and tests
- [x] Update QS and tests
- [x] Update integration tests
- [x] Unified tests for QS
- [x] Memstore support
- [x] Bolt support
- [x] LevelDB support (switch to protobuf)
- [x] Mongo support (untyped)
- [x] SQL support (untyped)
- [x] GAE support
- [x] Support for native types (int, float, etc)
- [x] Typed values in Mongo
- [x] Typed values in SQL
- [ ] Typed values in GAE
- [x] Path lib support
- [x] Gremlin support
- [x] Recognize TypedString and LandString in cquads parser
- [x] Registration for TypedString converters
- [ ] Support for custom types
- [ ] Migration code for SQL

#### Problems so far:
- [ ] goprotobuf will not work on GAE, thus we have no support for typed values for now

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/378)


dennwc:
Performance improvements for leveldb (d47e03d42637be8ef9c12685dcee4561958e6706):
```
benchmark                                     old ns/op     new ns/op     delta
BenchmarkNamePredicate-8                      908541        764269        -15.88%
BenchmarkLargeSetsNoIntersection-8            152631283     102180698     -33.05%
BenchmarkVeryLargeSetsSmallIntersection-8     370620517     249059029     -32.80%
BenchmarkHelplessContainsChecker-8            677294497     460773274     -31.97%
BenchmarkHelplessNotContainsFilms-8           660483822     436792429     -33.87%
BenchmarkHelplessNotContainsActors-8          688224013     467125431     -32.13%
BenchmarkNetAndSpeed-8                        21529498      15497605      -28.02%
BenchmarkKeanuAndNet-8                        10909581      8805261       -19.29%
BenchmarkKeanuAndSpeed-8                      14575400      11968996      -17.88%
BenchmarkKeanuOther-8                         89169973      62986376      -29.36%
BenchmarkKeanuBullockOther-8                  147350741     130021983     -11.76%
BenchmarkSaveBogartPerformances-8             4887044       3698388       -24.32%
```

dennwc:
### TODO list:
- [ ] GAE support
- [ ] Typed values in Mongo
- [ ] Typed values in SQL
- [ ] Path lib support
- [ ] Gremlin support
- [ ] MQL/SExp support?
- [ ] Recognize TypedString and LandString in cquads parser
- [ ] Registration for TypedString converters

dennwc:
Commit 6ae2eaccec0fd79fda87a2f5ce5e44f3bece3437 fixes #290. The syntax:
```js
g.V().Filter( gt(10) ).All()
```

kortschak:
Nice plan.

barakmich:
Reviewed 10 of 56 files at r1, 29 of 78 files at r2, 3 of 14 files at r3.
Review status: 40 of 95 files reviewed at latest revision, 4 unresolved discussions.

---

*[graph/bolt/all_iterator.go, line 45 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgbOAoVxzoNh7Tay4N:-KMgbOApl5vjeOW1GwW_:-1509546067) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/bolt/all_iterator.go#L45)):*
> ```Go
> func NewAllIterator(bucket []byte, d quad.Direction, qs *QuadStore) *AllIterator {
> 	return &AllIterator{
> 		nodes:  d == 0,
> ```

`d == quad.Any`?

---

*[graph/path/path.go, line 75 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgelIsQnJftWttqjBs:-KMgelIsQnJftWttqjBt:-1340118616) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/path/path.go#L75)):*
> ```Go
> 
> // StartMorphism creates a new Path with no underlying QuadStore.
> func StartMorphism(nodes ...quad.Value) *Path {
> ```

I'm of multiple minds about this. It breaks current API users. On the other hand, they're likely vendoring us. 

I'm open to suggestions, and not strictly against it as written, but we should at least discuss.  I'd lean toward making these `...interface{}` and type assert between strings and quad.Value for these cases, exactly as done in `morphism_apply_functions.go`

(Same also for the other ones in this change)

---

*[graph/proto/serializations.proto, line 66 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMggn94CsiCma5CeMVJ:-KMggn95tMtjo_Fb1s2r:726469523) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/proto/serializations.proto#L66)):*
> ```protobuf
>     int32 nanos = 2;
>   }
>   oneof value {
> ```

:clap: 

---

*[quad/value.go, line 62 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMjBzopEbUZHmK9lxDP:-KMjBzoqrwS-oedRHJSq:-10665999) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/quad/value.go#L62)):*
> ```Go
> type Raw string
> 
> func (s Raw) String() string { return string(s) }
> ```

After tracing down why the query language had to change --- `g.V('"Humphrey Bogart"')`, notice extra quotes, I came back here.

String() is being conflated with, effectively, MarshalRDF().

Sometimes these are the same. But in query languages, we'd rather interact with their string-ish representations, not their RDF-ish representations -- unless they *are* the same.

TL;DR, an RDF() method, used mainly in this package, and the String() function should work okay in the others.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

dennwc:
Review status: 40 of 95 files reviewed at latest revision, 3 unresolved discussions.

---

*[graph/path/path.go, line 75 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgelIsQnJftWttqjBs:-KMjRcHWp6j6Z0V9pE5G:-598674595) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/path/path.go#L75)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> I'm of multiple minds about this. It breaks current API users. On the other hand, they're likely vendoring us. 
> 
> I'm open to suggestions, and not strictly against it as written, but we should at least discuss.  I'd lean toward making these `...interface{}` and type assert between strings and quad.Value for these cases, exactly as done in `morphism_apply_functions.go`
> 
> (Same also for the other ones in this change)
</details>
This was a hard decision, in fact. `interface{}` is nice because you can seamlessly pass `string`, `int` and other values. But it also hides any potential errors. Also, each function must call some helper that casts values properly. I decided it's better to let user cast it once on his side and make our code cleaner. But I'm still in doubt.

---

*[quad/value.go, line 62 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMjBzopEbUZHmK9lxDP:-KMjK-zuNcgHDQxX_MQ7:131596061) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/quad/value.go#L62)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> After tracing down why the query language had to change --- `g.V('"Humphrey Bogart"')`, notice extra quotes, I came back here.
> 
> String() is being conflated with, effectively, MarshalRDF().
> 
> Sometimes these are the same. But in query languages, we'd rather interact with their string-ish representations, not their RDF-ish representations -- unless they *are* the same.
> 
> TL;DR, an RDF() method, used mainly in this package, and the String() function should work okay in the others.
</details>
Yes, I also like an idea to separate RDF marshaling/unmarshaling into it's own method set.

But the question regarding `g.V('"Humphrey Bogart"')` is more specific.  It works this way: any string is assumed to be in raw RDF (quad.Raw), thus it calculates hash of RDF form and then saves it in iterator. QuadStore will use hash to do lookup and it may hit quad.String node `Humphrey Bogart` (without quotes), because its RDF representation is the same as `"Humphrey Bogart"` (quad.Raw value). In this case queries like `g.V("<bob>")` and `g.V('"Bob"')` works, and they also works with helpers: `g.V(iri("bob"))`, `g.V(str("Bob"))`. I agree that this is too verbose. The alternative can be following: assume that strings are of quad.String type, and make helper for RDF form. In this case only `g.V(iri("bob"))` and `g.V(raw("<bob>"))` will work, but for strings it would be nicer: `g.V("Bob")` and `g.V(raw('"Bob"'))`.

And it'a also leads to broader discussion. Are nodes `10`, `"10"` and `"10"^^<int>` the same? What form should be hashed? May it use these forms interchangeably? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

barakmich:
*[quad/value.go, line 62 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMjBzopEbUZHmK9lxDP:-KMjcB--w6fs2MuB_RYc:1136560814) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/quad/value.go#L62)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary>
> Yes, I also like an idea to separate RDF marshaling/unmarshaling into it's own method set.
> 
> But the question regarding `g.V('"Humphrey Bogart"')` is more specific.  It works this way: any string is assumed to be in raw RDF (quad.Raw), thus it calculates hash of RDF form and then saves it in iterator. QuadStore will use hash to do lookup and it may hit quad.String node `Humphrey Bogart` (without quotes), because its RDF representation is the same as `"Humphrey Bogart"` (quad.Raw value). In this case queries like `g.V("<bob>")` and `g.V('"Bob"')` works, and they also works with helpers: `g.V(iri("bob"))`, `g.V(str("Bob"))`. I agree that this is too verbose. The alternative can be following: assume that strings are of quad.String type, and make helper for RDF form. In this case only `g.V(iri("bob"))` and `g.V(raw("<bob>"))` will work, but for strings it would be nicer: `g.V("Bob")` and `g.V(raw('"Bob"'))`.
> 
> And it'a also leads to broader discussion. Are nodes `10`, `"10"` and `"10"^^<int>` the same? What form should be hashed? May it use these forms interchangeably? 
</details>
Now, and going forward we'll need to think about the type space of JS and the type space of RDF and the representation of the latter in NQuads.

The way you write 10 in NQuads is `"10"^^<int>`. When strictly parsing quads, this becomes a `quad.Value(int(10))`. When represented in JS, this is `10`, of JS type number.

`"10"` in NQuads is `quad.Value(string("10"))` is "10" the JS string. Note this is different than the above: there is exactly one 10-the-int and one 10-the-string.

`"Humphrey Bogart"` in NQuads is `quad.Value(string("Humphrey Bogart")` is "Humphrey Bogart" the JS string.

Point is, we can choose the functions to represent a quad.Value (a typed RDF value) into both NQuads and into our JS environment.

Incidentally, we're creating functions to handle the w3c default types, but, should we want to extend our type space, we can even go so far as to declare our own IRIRef types. -- but that for another day.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

barakmich:
Reviewed 9 of 56 files at r1, 35 of 78 files at r2, 11 of 14 files at r3.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMk21DV8_ggbZsfrhgJ:-KMk21DWhL2GyFMz4yK8:792364673) (no related file):*
FYI; I completely understand the rewrite of the table structure under SQL. I even think it's the right change. I'm just a little curious how it might perform differently, though I'd imagine not much for the main query. The round-trips will kill you, however, resolving hashes to nodes. A follow up on this is to restore some performance by either (a) keeping an encoding on the table as there used to be (subject, predicate, object fields) and parsing them on the Cayley side when they come back, or (b) doing an explicit join on the top level.

---

*[graph/path/path.go, line 75 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgelIsQnJftWttqjBs:-KMjtX-XFQREr9K_wHiK:-1700465039) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/path/path.go#L75)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary>
> This was a hard decision, in fact. `interface{}` is nice because you can seamlessly pass `string`, `int` and other values. But it also hides any potential errors. Also, each function must call some helper that casts values properly. I decided it's better to let user cast it once on his side and make our code cleaner. But I'm still in doubt.
</details>
Yeah, I'm not sure. I'd like another thought or two

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

dennwc:
Review status: 87 of 95 files reviewed at latest revision, 3 unresolved discussions.

---

*[graph/bolt/all_iterator.go, line 45 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgbOAoVxzoNh7Tay4N:-KMkMALFvqOAQ6PbG-fF:-499146009) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/bolt/all_iterator.go#L45)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> `d == quad.Any`?
</details>
Done.

---

*[quad/value.go, line 62 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMjBzopEbUZHmK9lxDP:-KMkMHsOtdUNSh3leiLO:-19782824) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/quad/value.go#L62)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> Now, and going forward we'll need to think about the type space of JS and the type space of RDF and the representation of the latter in NQuads.
> 
> The way you write 10 in NQuads is `"10"^^<int>`. When strictly parsing quads, this becomes a `quad.Value(int(10))`. When represented in JS, this is `10`, of JS type number.
> 
> `"10"` in NQuads is `quad.Value(string("10"))` is "10" the JS string. Note this is different than the above: there is exactly one 10-the-int and one 10-the-string.
> 
> `"Humphrey Bogart"` in NQuads is `quad.Value(string("Humphrey Bogart")` is "Humphrey Bogart" the JS string.
> 
> Point is, we can choose the functions to represent a quad.Value (a typed RDF value) into both NQuads and into our JS environment.
> 
> Incidentally, we're creating functions to handle the w3c default types, but, should we want to extend our type space, we can even go so far as to declare our own IRIRef types. -- but that for another day.
</details>
Done. Gremlin now assumes that "<>" strings are IRI, "_:" are BNodes and everything else is a String.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

barakmich:
We're almost there. With the corresponding export for JS strings (in comments) and the updated tests to match, I think we can merge it!

---

Reviewed 8 of 8 files at r4.

---

*[query/gremlin/environ.go, line 234 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMphVEYlfQ4dcEpxd60:-KMphVEYlfQ4dcEpxd61:-788682789) ([raw file](https://github.com/cayleygraph/cayley/blob/0508015d04d590a9b6611223bdd4c21feb835304/query/gremlin/environ.go#L234)):*
> ```Go
> 	case string:
> 		if len(v) > 2 {
> 			if v[0] == '<' && v[len(v)-1] == '>' {
> ```

Maybe this could be a function in quad? `quad.FromJS(v)`? Because...

---

*[query/gremlin/session.go, line 222 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMphrkerdN-frVPBUzH:-KMphrkerdN-frVPBUzI:-574836393) ([raw file](https://github.com/cayleygraph/cayley/blob/0508015d04d590a9b6611223bdd4c21feb835304/query/gremlin/session.go#L222)):*
> ```Go
> 			for _, k := range tagKeys {
> 				if name := s.qs.NameOf(tags[k]); name != nil {
> 					obj[k] = name.String()
> ```

Here we can have name.ToJS() -- and remove the extraneous quotes on the returned values.

This is hugely important for things like

```
<v> <foo> "node"
<v> <name> "v"
```

```
s = g.V("node").In("<foo>").Out("<name>").ToValue
g.V(s) // s is '"v"' not 'v', and this query won't match the right node
```

In short, if we accept it, we have to return it under the same rules.

Fortunately, that function is easy. It's just:
* If we're of type quad.String
  * return internal string
* else return quad.String(), as this line does.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-:-KMpj7uYXE_J_B7XGNuJ:788313866)*

dennwc:
Review status: 91 of 95 files reviewed at latest revision, 3 unresolved discussions.

---

*[query/gremlin/environ.go, line 234 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMphVEYlfQ4dcEpxd60:-KMsdHeSrSYk0aYXKsIc:-1180466562) ([raw file](https://github.com/cayleygraph/cayley/blob/0508015d04d590a9b6611223bdd4c21feb835304/query/gremlin/environ.go#L234)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> Maybe this could be a function in quad? `quad.FromJS(v)`? Because...
</details>
I think it OK as is, because I hope that JS will be the only place which uses this. If we'll find it useful somewhere else I'll move it.

---

*[query/gremlin/session.go, line 222 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMphrkerdN-frVPBUzH:-KMsd-NgIP3nCS6gR59H:-1027399232) ([raw file](https://github.com/cayleygraph/cayley/blob/0508015d04d590a9b6611223bdd4c21feb835304/query/gremlin/session.go#L222)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> Here we can have name.ToJS() -- and remove the extraneous quotes on the returned values.
> 
> This is hugely important for things like
> 
> ```
> <v> <foo> "node"
> <v> <name> "v"
> ```
> 
> ```
> s = g.V("node").In("<foo>").Out("<name>").ToValue
> g.V(s) // s is '"v"' not 'v', and this query won't match the right node
> ```
> 
> In short, if we accept it, we have to return it under the same rules.
> 
> Fortunately, that function is easy. It's just:
> * If we're of type quad.String
>   * return internal string
> * else return quad.String(), as this line does.
</details>
Done. Checked every code path, now strings are not escaped in results.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

barakmich:
Reviewed 1 of 78 files at r2, 4 of 4 files at r5.

---

*[graph/path/path.go, line 75 \[r3\]](https://reviewable.io:443/reviews/cayleygraph/cayley/378#-KMgelIsQnJftWttqjBs:-KMu06Kmj-kSZhcD2PNw:5173665) ([raw file](https://github.com/cayleygraph/cayley/blob/b99aa1dab33f82f6f3a16d1f8f81e77492f73fd8/graph/path/path.go#L75)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary>
> Yeah, I'm not sure. I'd like another thought or two
</details>
OK

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/378)*

barakmich:
LGTM. Let's do this, the die is cast.

PR #394, opened 2016-06-17
graph/sql: add support for cockroachdb

d4l3k:
This fleshes out the multiple SQL flavor support in the graph/sql package. It
adds an `db_sql_flavor` option that can either be `postgres` (default) or
`cockroach`.

This also implements opts.IgnoreDup for postgres and cockroach.

The cockroach SQL dialect specific changes:
- Not using `WITH (FILLFACTOR = %d)` for indexes as it's not supported.
- Not using `pq.CopyIn` as `COPY <table> (<columns, ...) FROM STDIN` isn't
  supported yet.
- Retrying inserts when the "restart transaction" error is returned.
  https://www.cockroachlabs.com/docs/transactions.html#transaction-retries
#### Example config file using cockroach

``` json
{
  "database": "sql",
  "db_path": "postgres://root@localhost:26257/quads?sslmode=disable",
  "db_options": {
    "db_sql_flavor": "cockroach"
  },
  "read_only": false
}
```
#### Other Notes

It seems like that using a high load size when bulk inserting has a tendency to cause connection timeouts when inserting with cockroachdb. That may just because I'm running a multinode cluster on my resource limited laptop. Lowering it to 1000 seems to fix it.

``` bash
cayley load --config="cayley.json" --quads=data/30kmoviedata.nq.gz --load_size=1000
```

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -124,6 +124,16 @@ Optionally disable syncing to disk per transaction. Nosync being true means much
 
 The name of the database within MongoDB to connect to. Manages its own collections and indices therein.
 
+### SQL
+
+#### **`db_sql_flavor`**
+
+  * Type: String
+  * Default: "postgres"
+
+The SQL flavor to use with the sql driver. Can be "postgres" or "cockroach".
+
+
 ## Per-Replication Options
 
 The `replication_options` object in the main configuration file contains any of these following options that change the behavior of the replication manager.
--- a/graph/sql/quadstore.go
+++ b/graph/sql/quadstore.go
@@ -36,9 +36,18 @@ var (
 	hashSize = sha1.Size
 )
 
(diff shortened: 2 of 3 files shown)


dennwc:
@d4l3k Looks good overall. Does it pass integration tests? I mean tests in `/integration`, Travis fail is unrelated.

d4l3k:
For some reason I can't get the postgres integration tests to pass on either master or my fork. I'm using `postgres (PostgreSQL) 9.5.2`.

As for cockroach, I missed the part of the codebase that used joins so right now only very basic queries work.

dennwc:
@d4l3k Any update on this? HEAD was changed slightly, now PG tests should be OK.

dennwc:
Closing in favor of #489

PR #396, opened 2016-07-01
Use standard golang/glog instead of a fork

tzneal:
Using a fork causes issues when embedding cayley due to duplicate global
flags between glog and the fork.  This causes applications which have
dependencies on glog natively, or through another dependency to be
unable to use cayley without manually modifing its imports.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -10,8 +10,8 @@
 			"Rev": "9e5f7f4d07ca576562618c23e8abadda278b684f"
 		},
 		{
-			"ImportPath": "github.com/barakmich/glog",
-			"Rev": "fafcb6128a8a2e6360ff034091434d547397d54a"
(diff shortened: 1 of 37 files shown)


dennwc:
It's a more generic issue, actually. We should not force users to import particular log package. So it would be better to make a separate log sub-package with a Logger interface, so user may use whatever library he likes.

But thanks, anyway :)

tzneal:
While it would be nice to not have a dependency on glog, I think that a dependency on the standard glog vs a fork is a lesser evil.  

Currently we apply this CL to avoid a panic on startup due to duplicate flags between the fork and the real glog:

    panic:  flag redefined: log_dir

dennwc:
Yes, we had the same problem. Don't worry, I will change logging package shortly. And in any case glog will be the default.

tzneal:
Great, thanks!

PR #440, opened 2016-07-09
Don't iterate over all quads unless necessary

josephschorr:
Makes ApplyDeltas slightly faster when checking is not strictly necessary

### DIFF
--- a/graph/memstore/quadstore.go
+++ b/graph/memstore/quadstore.go
@@ -111,23 +111,25 @@ func newQuadStore() *QuadStore {
 }
 
 func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOpts) error {
-	// Precheck the whole transaction
-	for _, d := range deltas {
-		switch d.Action {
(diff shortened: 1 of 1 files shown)


graph/memstore/quadstore.go
```diff
@@ -111,23 +111,25 @@ func newQuadStore() *QuadStore {
...
-				if _, exists := qs.indexOf(d.Quad); !exists {
-					return graph.ErrQuadNotExist
+			case graph.Delete:
+				if !ignoreOpts.IgnoreMissing {
+					if _, exists := qs.indexOf(d.Quad); !exists {
+						return graph.ErrQuadNotExist
+					}
 				}
+			default:
+				return errors.New("memstore: invalid action")
```
dennwc: This return may be skipped now, and panic below will trigger. Replace panic with the same return.

josephschorr: Done

dennwc:
@josephschorr I think it's worth to review other backends too. Mongo at least might suffer from the same issue.

PR #443, opened 2016-07-10
graph: provide more meaningful errors for bad actions

kortschak:
This leads the way to a richer error handling that may allow the reduction of logging within packages.

### DIFF
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -305,9 +305,8 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 		clog.Errorf("Couldn't write to DB for Delta set. Error: %v", err)
 		qs.horizon = oldHorizon
 		qs.size = oldSize
-		return err
 	}
-	return nil
+	return err
 }
 
 func (qs *QuadStore) buildQuadWrite(tx *bolt.Tx, q quad.Quad, id int64, isAdd bool) error {
--- a/graph/bolt/quadstore.go
+++ b/graph/bolt/quadstore.go
@@ -253,16 +253,16 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 		sizeChange := int64(0)
 		for _, d := range deltas {
 			if d.Action != graph.Add && d.Action != graph.Delete {
-				return errors.New("bolt: invalid action")
+				return &graph.DeltaError{Delta: d, Err: graph.ErrInvalidAction}
 			}
 			p := deltaToProto(d)
 			bytes, err := p.Marshal()
 			if err != nil {
-				return err
+				return &graph.DeltaError{Delta: d, Err: err}
 			}
 			err = b.Put(qs.createDeltaKeyFor(d.ID.Int()), bytes)
 			if err != nil {
-				return err
+				return &graph.DeltaError{Delta: d, Err: err}
 			}
 		}
 		for _, d := range deltas {
@@ -274,7 +274,7 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 				if err == graph.ErrQuadNotExist && ignoreOpts.IgnoreMissing {
 					continue
 				}
-				return err
+				return &graph.DeltaError{Delta: d, Err: err}
 			}
 			delta := int64(1)
 			if d.Action == graph.Delete {
--- a/graph/memstore/quadstore.go
+++ b/graph/memstore/quadstore.go
@@ -15,7 +15,6 @@
 package memstore
 
 import (
-	"errors"
 	"time"
 
 	"github.com/cayleygraph/cayley/clog"
@@ -118,17 +117,17 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 			case graph.Add:
 				if !ignoreOpts.IgnoreDup {
 					if _, exists := qs.indexOf(d.Quad); exists {
-						return graph.ErrQuadExists
+						return &graph.DeltaError{Delta: d, Err: graph.ErrQuadExists}
 					}
 				}
 			case graph.Delete:
 				if !ignoreOpts.IgnoreMissing {
 					if _, exists := qs.indexOf(d.Quad); !exists {
-						return graph.ErrQuadNotExist
+						return &graph.DeltaError{Delta: d, Err: graph.ErrQuadNotExist}
 					}
 				}
 			default:
-				return errors.New("memstore: invalid action")
+				return &graph.DeltaError{Delta: d, Err: graph.ErrInvalidAction}
 			}
 		}
 	}
@@ -147,7 +146,7 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 				err = nil
 			}
 		default:
-			return errors.New("memstore: invalid action")
+			err = &graph.DeltaError{Delta: d, Err: graph.ErrInvalidAction}
 		}
 		if err != nil {
 			return err
@@ -193,7 +192,7 @@ func (qs *QuadStore) indexOf(t quad.Quad) (int64, bool) {
 
 func (qs *QuadStore) AddDelta(d graph.Delta) error {
 	if _, exists := qs.indexOf(d.Quad); exists {
-		return graph.ErrQuadExists
+		return &graph.DeltaError{Delta: d, Err: graph.ErrQuadExists}
 	}
 	qid := qs.nextQuadID
 	qs.log = append(qs.log, LogEntry{
@@ -226,7 +225,7 @@ func (qs *QuadStore) AddDelta(d graph.Delta) error {
 func (qs *QuadStore) RemoveDelta(d graph.Delta) error {
 	prevQuadID, exists := qs.indexOf(d.Quad)
 	if !exists {
-		return graph.ErrQuadNotExist
+		return &graph.DeltaError{Delta: d, Err: graph.ErrQuadNotExist}
 	}
 
 	quadID := qs.nextQuadID
--- a/graph/mongo/quadstore.go
+++ b/graph/mongo/quadstore.go
@@ -227,11 +227,11 @@ func (qs *QuadStore) updateLog(d graph.Delta) error {
 	return err
 }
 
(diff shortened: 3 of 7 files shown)


graph/memstore/quadstore.go
```diff
@@ -117,17 +116,17 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 		case graph.Add:
 			if !ignoreOpts.IgnoreDup {
 				if _, exists := qs.indexOf(d.Quad); exists {
-					return graph.ErrQuadExists
+					return &graph.QuadError{Action: d.Action, Quad: d.Quad, Err: graph.ErrQuadExists}
```
dennwc: I think this is a bad idea - caller may already compare error with ErrQuadExists. Why to wrap it?

I thinks it's better to remove definition of ErrQuadExists as a variable and make it an error type with single Quad field. This will break code, but at least will show to the user that some changes are necessary.

graph/memstore/quadstore.go
```diff
@@ -117,17 +116,17 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
...
 		case graph.Delete:
 			if !ignoreOpts.IgnoreMissing {
 				if _, exists := qs.indexOf(d.Quad); !exists {
-					return graph.ErrQuadNotExist
+					return &graph.QuadError{Action: d.Action, Quad: d.Quad, Err: graph.ErrQuadNotExist}
 				}
 			}
 		default:
-			return errors.New("memstore: invalid action")
+			return &graph.QuadError{Action: d.Action, Quad: d.Quad, Err: graph.ErrInvalidAction}
```
dennwc: Make it more compact:
```Go
type ErrInvalidAction struct {
   Delta graph.Delta
}
```

graph/memstore/quadstore.go
kortschak: There are three bad actions. Doing it that way would either require having three distinct error types or alternatively the error string of QuadError could be dependent on the Action field: +1 => "quad exists", -1 => "quad does not exist", other => "invalid".

The approach you are suggesting was what I originally was considering (actually the even simpler approach of `type QuadError quad.Quad`). The approach here mimics os.PathError.

kortschak: The other thing is that this approach allows more informative errors to be placed in the Err field. For example in bolt the actual cause for a failure could be included in the Err field using something along the lines of https://github.com/go-errgo/errgo rather than logging.

graph/memstore/quadstore.go
kortschak: Yeah, I also was thinking about that, but I don't think the Delta ID or Timestamp fields are wanted here.

graph/memstore/quadstore.go
dennwc: Generic quad error type is ok, but we definitely have 3 concrete errors which deserves separate types.

graph/memstore/quadstore.go
dennwc: They may be useful for the caller. At least he will not be forced to search for the right Delta by hand.

graph/memstore/quadstore.go
kortschak: Why? They are all errors which describe failures in delta application. This is almost entirely analogous to the use of os.PathError.

If we document that methods return a QuadError on failure then callers can assert if they want to find the error.

dennwc: It will silently break code right now. All tests that can show this are in #378.

PathError also has some helpers like IsNotExists that makes a life easier. For now you are forcing a caller to do 2 type asserts in a row.

kortschak: Two? At the most there is one, error asserted to QuadError, the Err field value is comparable.

graph/memstore/quadstore.go
kortschak: Done.

graph/memstore/quadstore.go
dennwc: Ah, sorry, was thinking about other implementation.

In any case, these points remains:
a) It silently breaks callers code
b) Caller is forced to type assert AND compare instead of one compare / type assertion / calling helper
c) Generic quad error does not need to know about Delta's Action

kortschak: 1) and 2) can be fixed by unexporting the errors that exist (note that no references to these values exist in the code base as it stands) and providing helpers as exist for os: IsQuadExist(error) bool and IsQuadNotExist(error) bool. 3) is addressed by refashioning the error to be a DeltaError (which was what I was really intending).

graph/memstore/quadstore.go
```diff
@@ -193,7 +193,7 @@ func (qs *QuadStore) indexOf(t quad.Quad) (int64, bool) {
 
 func (qs *QuadStore) AddDelta(d graph.Delta) error {
 	if _, exists := qs.indexOf(d.Quad); exists {
-		return graph.ErrQuadExists
+		return &graph.DeltaError{Delta: d, Err: graph.ErrQuadExists}
```
dennwc: These could be without wrapping. There is only one Delta in any case.

kortschak: That would complexify the code and the documentation. With that we get variance on the types of errors that the QuadWriter type delta methods return.

graph/memstore/quadstore.go
dennwc: 1,2) Where was no references because nobody cared about checking this behavior for some reason. #378 does this. Maybe some apps also do. Unexporting sounds good to signal that something changed to the user, but it also means that backends could not use the error.
3) Looks good this way.

graph/memstore/quadstore.go
dennwc: Actually this method is not even in QuadStore interface. Just memstore addition.

graph/memstore/quadstore.go
kortschak: >Unexporting sounds good to signal that something changed to the user, but it also means that backends could not use the error.

Eek. Yes.

graph/bolt/quadstore.go
```diff
@@ -274,7 +274,7 @@ func (qs *QuadStore) ApplyDeltas(deltas []graph.Delta, ignoreOpts graph.IgnoreOp
 				if err == graph.ErrQuadNotExist && ignoreOpts.IgnoreMissing {
 					continue
 				}
-				return err
+				return &graph.DeltaError{Delta: d, Err: err}
```
dennwc: Does this means that ApplyDeltas should return DeltaError in most cases? There are multiple brunches in this function that doesn't follow this behavior.

graph/memstore/quadstore.go
kortschak: QuadWriter interface, but yes, this gets called by ApplyDeltas. I think it's a relic, rather than an addition, the QuadWriter approach came after the simple Quad addition.

dennwc: Nope, ApplyDeltas does on it's own. So I think unwrapping is ok here.

kortschak: Call at line 140.

dennwc: My bad :) In any case 150, 153 and other branches around will need wrapping too.

graph/bolt/quadstore.go
kortschak: It seems that it's not possible to do that. That is unfortunate.

dennwc: At least 261 and 265 are both in delta loop. Worth changing.

kortschak: Done.

graph/memstore/quadstore.go
kortschak: AddDelta and RemoveDelta already have the error wrapped. The invalid action does though.

graph/mongo/quadstore.go
```diff
@@ -268,7 +267,7 @@ func (qs *QuadStore) ApplyDeltas(in []graph.Delta, ignoreOpts graph.IgnoreOpts)
 	for _, d := range in {
 		err := qs.updateQuad(d.Quad, d.ID.Int(), d.Action)
 		if err != nil {
-			return err
+			return &graph.DeltaError{Delta: d, Err: err}
```
dennwc: 264 still unwrapped

dennwc:
Please add these helpers also.

graph/mongo/quadstore.go
kortschak: Thanks.

kortschak:
PTAL

graph/quadwriter.go
```diff
@@ -59,13 +69,50 @@ func (h *Handle) Close() {
...
+
+func (e *DeltaError) Error() string {
+	return e.Delta.Action.String() + " " + e.Delta.Quad.String() + ": " + e.Err.Error()
+}
+
+// IsQuadExist returns whether an error is a DeltaError
+// with the Err field equalt to ErrQuadExists.
+func IsQuadExist(err error) bool {
+	de, ok := err.(*DeltaError)
+	return ok && de == ErrQuadExists
```
dennwc: Add direct comparison also

graph/quadwriter.go
```diff
@@ -59,13 +69,50 @@ func (h *Handle) Close() {
...
+// IsQuadExist returns whether an error is a DeltaError
+// with the Err field equalt to ErrQuadNotExist.
+func IsQuadNotExist(err error) bool {
+	de, ok := err.(*DeltaError)
+	return ok && de == ErrQuadNotExist
+}
+
+// IsQuadExist returns whether an error is a DeltaError
+// with the Err field equalt to ErrInvalidAction.
+func IsInvalidAction(err error) bool {
```
dennwc: Might not be necessary. But it's ok to have it if you think it will be useful widely.

graph/quadwriter.go
```diff
@@ -59,13 +69,50 @@ func (h *Handle) Close() {
...
+}
+
+// IsQuadExist returns whether an error is a DeltaError
+// with the Err field equalt to ErrQuadExists.
+func IsQuadExist(err error) bool {
+	de, ok := err.(*DeltaError)
+	return ok && de == ErrQuadExists
+}
+
+// IsQuadExist returns whether an error is a DeltaError
```
dennwc: Name is wrong in comments

kortschak: I need to get some rest. Thanks.

dennwc:
@kortschak Thank you for the change! Current error handling and logging is a bit messy.

kortschak:
It's something I've been meaning to fix for almost two years (#157). This is a step in the right direction hopefully.

PR #444, opened 2016-07-13
Add Clone() to Path for easier reuse of Path segments

josephschorr:
This allows for construction of sub-Paths, and then cloning to use them under multiple `And`s or `Or`s.

### DIFF
--- a/graph/path/path.go
+++ b/graph/path/path.go
@@ -98,6 +98,16 @@ func NewPath(qs graph.QuadStore) *Path {
 	}
 }
 
+// Clone returns a clone of the current path.
+func (p *Path) Clone() *Path {
+	stack := p.stack
+	return &Path{
+		stack:       stack[:len(stack):len(stack)],
+		qs:          p.qs,
+		baseContext: p.baseContext,
+	}
+}
+
 // Reverse returns a new Path that is the reverse of the current one.
 func (p *Path) Reverse() *Path {
 	newPath := NewPath(p.qs)


dennwc:
You should copy the stack in Clone

dennwc:
Even a better way: limit capacity of saved stack

josephschorr:
@dennwc: Any reason why? As far as I can tell, all the modifiers on the Path will call `append`, so having it be the same slice should be safe.

dennwc:
@josephschorr I don't think so. Consider this example:
```Go
p := path.Start("node") // slice len is 1, cap 1

p = p.Out("foo") // slice len is 2, cap is ~10
p2 := p.Clone() // same slice

p = p.Out("bar") // len 3, cap ~10
p2 = p2.In("baz") // same slice - overriding previous action
```

kortschak:
@dennwc Yes, not using cap limits in contexts like this has given us subtle and difficult to debug errors in gonum/graph.

josephschorr:
@kortschak: Such as? Just to fulfill my own curiosity.

@dennwc: Should I do a straight copy, or copy to a cap-limited slice?

kortschak:
https://github.com/gonum/graph/issues/61 with the fix [here](https://github.com/gonum/graph/pull/62).

josephschorr:
@dennwc: Change made. Please let me know if you had something else in mind.

dennwc:
Looks good. Thanks!

PR #449, opened 2016-07-17
GraphQL support

### LINKED ISSUE #370: Implement GraphQL
Please see it:
https://github.com/graphql-go/graphql

dennwc:
Adds support  for GraphQL query language. Resolves #370.

Warning: work in progress, should not be merged yet.

**TODO:**
- [x] query support
- [x] UI integration
- [x] resolve predicate naming issue
- [ ] implement fragments
- [ ] implement mutations?

### DIFF
--- a/appengine/appengine.go
+++ b/appengine/appengine.go
@@ -31,6 +31,7 @@ import (
 	_ "github.com/cayleygraph/cayley/writer"
 
 	// Register supported query languages
+	_ "github.com/cayleygraph/cayley/query/graphql"
 	_ "github.com/cayleygraph/cayley/query/gremlin"
 	_ "github.com/cayleygraph/cayley/query/mql"
 )
--- a/cmd/cayley/cayley.go
+++ b/cmd/cayley/cayley.go
@@ -56,6 +56,7 @@ import (
 
 	// Load supported query languages
 	_ "github.com/cayleygraph/cayley/query/gizmo"
+	_ "github.com/cayleygraph/cayley/query/graphql"
 	_ "github.com/cayleygraph/cayley/query/gremlin"
 	_ "github.com/cayleygraph/cayley/query/mql"
 	_ "github.com/cayleygraph/cayley/query/sexp"
--- a/glide.lock
+++ b/glide.lock
@@ -138,4 +138,6 @@ imports:
   - sasl
 - name: github.com/go-sql-driver/mysql
   version: 2e00b5cd70399450106cec6431c2e2ce3cae5034
+- name: github.com/dennwc/graphql
+  version: fd5c4aa13f2119414084f229600de0d73f174436
 testImports: []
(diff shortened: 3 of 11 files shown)


abourget:
Woah..  What's the state of this? That would be awesome!

dennwc:
@abourget The state is "it works", but I need someone to test it out.

Also it would be great to have a list of must-have GraphQL features that should be implemented.

robertmeta:
I think I am going to take this for a spin, even without fragment and mutations it is useful to us.

evseevnn:
How soon will be released?

dennwc:
Rebased on top of current master, and made some final modifications. Will merge shortly.

dan-f:
How are you addressing the fact that [URIs cannot be parsed as names in GraphQL](https://facebook.github.io/graphql/#sec-Names)?

dennwc:
@dan-f This is not strictly a GraphQL, so currently it uses a modified parser to handle IRIs and bnodes.

steffansluis:
@dennwc Although the modified parser allows a bit more flexibility than the standard GraphQL syntax, it's still not able to process all valid IRI's, such as `http://www.w3.org/1999/02/22-rdf-syntax-ns#type`, which breaks on the `-` in the IRI.

PS: Really like the GraphQL API! :tada::+1:

dennwc:
@steffansluis Thanks :) I'll open an issue to track parser-related problems. I'll also send a quick fix soon.

Also, as a notice, you can use `rdf:type` and it will be expanded to a full URL automatically.

steffansluis:
@dennwc Thank you for the quick fix, but now `http://www.w3.org/1999/02/22-rdf-syntax-ns#type` is breaking on the `#` and using `rdf:type` doesn't seem to work. I've tested with the Gizmo api, the following query works fine, resulting in a multitude of node id's as a result:
```
g.V("<http://schema.org/Thing>")
	.In("<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>")
	.All()
```

Whereas the GraphQL equivalent breaks:
```
{
  nodes(id: "<http://schema.org/Thing>") {
    id
    type: <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> @rev {
  		id
  	}
  }
}
```
with the error:
`{"data":null,"errors":[{"message":"Syntax Error GraphQL (8:1) Unexpected }\n\n7:   }\n8: }\n   ^\n","locations":null}]}`

Using `rdf:type` instead (with or without `<>`), I just get a null result for the type: `{"data":{"nodes":{"id":"http://schema.org/Thing","type":null}}}`.

stereobooster:
> This is not strictly a GraphQL, so currently it uses a modified parser to handle IRIs and bnodes.

I think we should name it GraphQL than. It can be confusing

Have you this one https://docs.dgraph.io/query-language/, but the name is horrible "GraphQL+-" (ungoogleable).

PR #453, opened 2016-07-23
Example of BoltDB

oren:


robertmeta:
@oren while you are in there -- can you remove the double close from this (and the other examples, and the quickstart doc) -- closing the nexter closes the iterator, so we only need to do one.

robertmeta:
One last thing -- you should use tempfile rather than a fixed path to the DB: https://golang.org/pkg/io/ioutil/#TempFile

PR #456, opened 2016-08-01
Fix Docker building

allanino:
We can see in https://quay.io/repository/cayleygraph/cayley that recent builds are failing.

I found out that there is a missing dependency, so this fix amounts to adding it to [Godeps/Godeps.json](https://github.com/cayleygraph/cayley/blob/master/Godeps/Godeps.json).

I also updated the Dockerfile to use relative imports, as this improves readability.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,12 +6,12 @@ WORKDIR /go/src/github.com/cayleygraph/cayley
 
 # Restore vendored dependencies
 RUN go get github.com/tools/godep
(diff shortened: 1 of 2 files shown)


robertmeta:
Sorry for the delay in getting to this -- but it looks good at first glance.

dennwc:
My bad, forgot to add it to Godep. Thanks for the fix @allanino !

PR #457, opened 2016-08-16
graph: fix Is*(err error) checkers to properly compare against *DeltaError

kris-runzer:
The three error checkers in **graph** were missing the `.Err` for their final comparisons.


barakmich:
LGTM

PR #459, opened 2016-08-17
user/pass functionality for mongo backend

dwhitena:
# What this PR does:
- allows for multiple Mongo hosts in options to Mongo backend
- allows for username in options to Mongo backend
- allows for password in options to Mongo backend

# How to test this PR:
- `go test` from `graph/mongo`

### DIFF
--- a/graph/mongo/quadstore.go
+++ b/graph/mongo/quadstore.go
@@ -17,6 +17,7 @@ package mongo
 import (
 	"encoding/hex"
 	"fmt"
+	"strings"
 	"time"
 
 	"gopkg.in/mgo.v2"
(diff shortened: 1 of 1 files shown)


graph/mongo/quadstore.go
```diff
@@ -113,38 +114,57 @@ func ensureIndexes(db *mgo.Database) error {
...
 	conn.SetSafe(&mgo.Safe{})
+	db := conn.DB("")
+	return ensureIndexes(db)
+}
+
+func dialMongo(addr string, options graph.Options) (*mgo.Session, error) {
+	var dialInfo mgo.DialInfo
+	dialInfo.Addrs = strings.Split(addr, ",")
+	user, ok, err := options.StringKey("username")
+	switch {
```
dennwc: Why not if-elseif?

graph/mongo/quadstore.go
```diff
@@ -113,38 +114,57 @@ func ensureIndexes(db *mgo.Database) error {
...
 
 func createNewMongoGraph(addr string, options graph.Options) error {
-	conn, err := mgo.Dial(addr)
+	conn, err := dialMongo(addr, options)
 	if err != nil {
 		return err
 	}
 	defer conn.Close()
 	conn.SetSafe(&mgo.Safe{})
+	db := conn.DB("")
```
dennwc: This is for default DB, I assume?

barakmich: Yeah, I double-checked that

graph/mongo/quadstore.go
dwhitena: My understanding is that the switch is more idiomatic.  "It's therefore possible—and idiomatic—to write an if-else-if-else chain as a switch" https://golang.org/doc/effective_go.html#switch

graph/mongo/quadstore.go
dwhitena: Yes, that will just use the DB specified in DialInfo (i.e., the one specified as default or the one explicitly specified in the options).

graph/mongo/quadstore.go
dennwc: Haven't seen that kind of code style in many project, actually. Also, the rest of the code in repo uses if-else-if for small condition chains. Can you replace these switches to match?

kortschak: Agree. In some cases in stdlib switches are used in small condition chains, but these generally refer to things that would be considered sets of cases on the switch expressionrather than what we see here. There doesn't even need to be if else, since the failing error case returns (can it return a nil?), then you do the work if ok is true.

dwhitena: Updated in the latest diff to if statements.  Also, @kortschak I think we need to return the empty struct rather than nil in this case.

dennwc: @dwhitena Nil is definitely better. Why an empty struct?

graph/mongo/quadstore.go
```diff
@@ -128,25 +128,25 @@ func dialMongo(addr string, options graph.Options) (*mgo.Session, error) {
...
 	user, ok, err := options.StringKey("username")
-	switch {
-	case err != nil:
+	if err != nil {
 		return &mgo.Session{}, err
-	case ok:
+	}
+	if ok {
 		dialInfo.Username = user
 		password, ok, err := options.StringKey("password")
```
dennwc: Should it be inside "username present" if statement? User may specify just a password. Or it will not work for Mongo?

graph/mongo/quadstore.go
dwhitena: @dennwc Yeah, forgot we had a pointer here instead of returning the struct itself.  Thanks.

barakmich:
LGTM.

PR #460, opened 2016-08-18
Glide migration

### LINKED ISSUE #451: Glide migration
We will be migrating from Godeps to Glide... Soon!

dennwc:
Closes #451


dennwc:
@barakmich May I drop Go 1.5 from the build matrix? It seems that vendoring is not working properly there. At least `go list` leads to some strange results.

barakmich:
You'll also want to commit glide.lock -- the yaml should tell which packages, and the lock which versions. See Torus for an example :)

dennwc:
So I need to drop all commit hashes from glide.yml and commit lock file instead. Correct?

barakmich:
Correct. Then, bumping current versions goes in the lock file, tracking (or locking for a good reason) a dep goes in the yaml

barakmich:
LGTM on green

PR #462, opened 2016-08-23
graph: replace UID method on Iterator with reflect-based

dennwc:


kortschak:
What's the rationale for this?

graph/iterator.go
```diff
@@ -333,35 +331,53 @@ func DumpStats(it Iterator) StatsContainer {
...
 		} else {
-			clog.Infof("%s %d NEXT DONE", strings.ToUpper(it.Type().String()), it.UID())
+			clog.Infof("%s %d NEXT DONE", strings.ToUpper(it.Type().String()), UID(it))
 		}
 	}
 	return ok
 }
+
+// UID returns the unique identifier of the iterator.
+func UID(it Iterator) uint64 {
```
kortschak: I think this is a bad idea for two reasons. The first is that Iterator is an interface which could conceivably be implemented by a user embedding a pointer type in a non-pointer type. This would break this approach to getting an ID. The second is that this approach to getting an identifier value is unsafe; at the moment for most cases it will work and the uses we have here are only informational, however other people may use the values as data rather than logging info - and the basis values may change. This doesn't happen often in the current compiler implementation, but the Go team are leaving open the option to implement a moving GC (and even in the current implementation, I have seen cases where using a pointer-based uid results in collisions due to address reuse).

dennwc: Good point! Totally forgot about moving GC, which makes current approach useless. Will try another approach then. Thanks!

PR #466, opened 2016-08-29
Clone receiver paths

wayeast:
This is a follow-up to a [previous issue](https://discourse.cayley.io/t/should-traversal-methods-change-their-receiver-path/308).

Said issue reported on path variables in Gremlin/javascript changing as new
paths were created as extensions from them.  This issue was resolved with
a subsequent [PR](https://github.com/cayleygraph/cayley/commit/cf824fe4bb9b5f64296c320f06febc3c33dbc9e0).

The problem, though, lies deeper.  Using cayley as a library and executing
a similar query pattern, the same results appear:

```
func main() {
     store, _ := cayley.NewGraph("leveldb", "/path/to/leveldb.db", nil)

     alice_path := path.StartPath(store, quad.IRI("alice"))
     alice_pre, _ := alice_path.Iterate(nil).AllValues(nil)
     fmt.Println("alice:", alice_pre)    // alice is <alice>

     alice_follows_path := alice_path.Out(quad.IRI("follows"))
     alice_follows, _ := alice_follows_path.Iterate(nil).AllValues(nil)
     fmt.Println("alice_follows:", alice_follows)    // alice follows <bob>

     alice_post, _ := alice_path.Iterate(nil).AllValues(nil)
     fmt.Println("alice:", alice_post)    // alice has turned into <bob>
}
```

Furthermore, for reasons I couldn't track down, I think this behavior contributed
to a pretty serious bug with the Except method on *Path.  Try the following and
watch what happens to your machine's memory usage:

```
func main() {
     fmt.Println("test_query2.go, building store")
     store, _ := cayley.NewGraph("leveldb", "/path/to/leveldb.db", nil)
     p := path.NewPath(store)
     coolpeople := p.Has(quad.IRI("status"), quad.String("cool_person"))
     bobfollowers := p.Has(quad.IRI("follows"), quad.IRI("bob"))
     reduced := bobfollowers.Except(coolpeople)
     vals, _ := reduced.Iterate(nil).AllValues(nil)
     fmt.Println("bob's followers who are not cool:", vals)
}
```

This commit includes a test in graph/path/path_test.go.  It does not undo the
changes to query/gremlin/traversals.go from the above-mentioned commit.
This probably means that there is some extra cloning that does not need to
happen and which should be removed.  The gremlin tests pass with this commit,
however.

### DIFF
--- a/graph/path/path.go
+++ b/graph/path/path.go
@@ -130,21 +130,24 @@ func (p *Path) Reverse() *Path {
 // Is declares that the current nodes in this path are only the nodes
 // passed as arguments.
 func (p *Path) Is(nodes ...quad.Value) *Path {
-	p.stack = append(p.stack, isMorphism(nodes...))
-	return p
+	np := p.Clone()
+	np.stack = append(np.stack, isMorphism(nodes...))
+	return np
 }
 
 // Filter represents the nodes that are passing comparison with provided value.
 func (p *Path) Filter(op iterator.Operator, node quad.Value) *Path {
-	p.stack = append(p.stack, cmpMorphism(op, node))
-	return p
+	np := p.Clone()
+	np.stack = append(np.stack, cmpMorphism(op, node))
+	return np
 }
 
 // Tag adds tag strings to the nodes at this point in the path for each result
 // path in the set.
 func (p *Path) Tag(tags ...string) *Path {
-	p.stack = append(p.stack, tagMorphism(tags...))
-	return p
+	np := p.Clone()
+	np.stack = append(np.stack, tagMorphism(tags...))
+	return np
 }
 
 // Out updates this Path to represent the nodes that are adjacent to the
@@ -157,8 +160,9 @@ func (p *Path) Tag(tags ...string) *Path {
 //  // to "F" labelled "follows".
 //  StartPath(qs, "A").Out("follows")
 func (p *Path) Out(via ...interface{}) *Path {
-	p.stack = append(p.stack, outMorphism(nil, via...))
-	return p
+	np := p.Clone()
+	np.stack = append(np.stack, outMorphism(nil, via...))
+	return np
 }
 
 // In updates this Path to represent the nodes that are adjacent to the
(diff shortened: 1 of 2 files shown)


dennwc:
Actually I expected user to clone path manually to use it later. But maybe this extra step is not expected by users. Let me check if we can do it without extra copying.

dennwc:
Ok, so here's the thing. According to [path.go:112](https://github.com/cayleygraph/cayley/pull/466/commits/e78aa8795ed9e3826a66b2ec56394d72771e2be2#diff-406f7535cba98fb5f76ea8814779f5abL112) Clone will return a new path with capacity-limited morphisms stack. In your case it will lead to unnecessary copy of stack on each path modification.

Clone cannot be changed to return non-limited stack because user may call it multiple times. Thus, we need a new unexported clone method that assumes that cloned path will be appended instead of current one. This method should replace Clone that is called in this PR. Maybe we can even remove original Clone later.

wayeast:
@dennwc Well, I guess I thought I was fixing something that was more clearly a bug than it is/was.  Your assumption that the user would clone a path manually is one that had not occurred to me, though thinking about it now it makes sense.  I have some thoughts on the question that are probably not going to be articulated well right now, but will throw them out anyway as food for thought.  Also, I'm sure that there is a theoretical language that one can learn in the university to describe very concisely what I will bumble here, but I don't possess the training to use it dexterously.  Many apologies for speaking in examples and analogies.

First, the position from which I made the changes in this PR.  To my mind, the most sensible thing is for a variable assignment operation to create on the left a separate and distinct value from what is on the right.  Eg. in the statement ```alice_follows_path := alice_path.Out(quad.IRI("follows"))```, ```alice_path``` remains the same ```alice_path``` as before, and ```alice_follows_path``` is a value which can be manipulated/extended without affecting ```alice_path```.  The changes in this PR are playing to this assumption.

I can see, though, the merit of your assumption -- that graph traversal methods do not represent new and distinct paths but rather modifications to an original path.  If we were to preserve this base assumption, I would suggest that 1) it should be made clear in documentation that traversal methods should be used for their _side effects_ and not their return values, and 2) this rule be enforced by having traversal methods not return any value at all, or at the least only return an eventual error value.  This way, if a user tries to use traversal methods in the way that I was doing in the examples above, they would know very quickly that something is wrong with their code.  As it is currently, it is too easy for a user to go blithely along with a completely mistaken notion of what their code is doing, and it would take (as it did in my case) some careful tracing of steps to figure out what the problem is.

I have been doing some messing about today with this second suggestion (removing return values from traversal methods).  The alice example above works, but such a change would involve some more in-depth modifications to the code base, not to mention of any code that users have already written and would expect to still run on new versions of cayley.  Next step will be to experiment with the unexported clone method you describe, but this will be for tomorrow (or asap).  In the meantime, whatever thoughts or dictums you might have on the proper direction would be appreciated.

wayeast:
@dennwc I've added a commit to this PR that, to the best of my understanding, addresses your concern about non-limited stacks.  See if this is in line with what you were thinking.  Thanks.

dennwc:
I had something like this in mind:
```
func (p *Path) clone() *Path {
	stack := p.stack
        // limit stack capacity of an old path, so append will not affect a new path's stack and will cause a silent copy
        p.stack = stack[:len(stack):len(stack)]
	return &Path{
		stack:       stack, // we expecting new path to be appended again
		qs:          p.qs,
		baseContext: p.baseContext,
	}
}
```
LGTM after this change :)

wayeast:
@dennwc changes made.

dennwc:
Thanks, merging!

PR #468, opened 2016-08-31
Support for RethinkDB

mikaelcabot:
Implementation is **heavily** inspired by the MongoDB implementation.
Initial (and simple) testing seems to be ok.


dennwc:
That's great! Will try it out soon. Does integration test passes, btw?

---

Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 24 unresolved discussions.

---

*[graph/rethinkdb/indexed_linksto.go, line 19 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV1KGyFTckiGPw2IpO:-KQV1KGzdOXAIRCZE4yj:bf8naiu) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/indexed_linksto.go#L19)):*
> ```Go
> }
> 
> // LinksTo is a MongoDB-dependent version of a LinksTo iterator. Like the normal
> ```

Remove references to Mongo in docs

---

*[graph/rethinkdb/indexed_linksto.go, line 226 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVAdGSOSbCacfTBXci:-KQVAdGSOSbCacfTBXcj:bqmjh3b) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/indexed_linksto.go#L226)):*
> ```Go
> 		it.nextIt.Close()
> 	}
> 	it.nextIt = nil
> ```

Also reset it.err

---

*[graph/rethinkdb/iterator.go, line 100 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVBHoXeY52d7ALEJi5:-KQVBHoXeY52d7ALEJi6:bqmjh3b) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/iterator.go#L100)):*
> ```Go
> func (it *Iterator) Reset() {
> 	it.Close()
> 	it.iter = it.makeRDBIterator()
> ```

Also reset it.err

---

*[graph/rethinkdb/quadstore.go, line 24 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV1exi3mbeqrLMBZbm:-KQV1exi3mbeqrLMBZbn:b-bc30fb) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L24)):*
> ```Go
> const nodeTableName = "nodes"
> const quadTableName = "quads"
> const logTableName = "log"
> ```

Merge into one const block (or one for exported and one for unexported)

---

*[graph/rethinkdb/quadstore.go, line 36 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV27X_ab-AwA144cIA:-KQV27XaG-BFdnFs5ADs:b-j935bu) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L36)):*
> ```Go
> }
> 
> type NodeHash string
> ```

Is it possible to store `[]byte` in Rethink without encoding overhead? If yes it's better to store these hashes without hex encoding, and type should be `[quad.HashSize]byte` in this case. It will matter for overall DB size.

---

*[graph/rethinkdb/quadstore.go, line 38 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV2-J6-GowzwZiHyOq:-KQV2-J6-GowzwZiHyOr:b-k01iwr) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L38)):*
> ```Go
> type NodeHash string
> 
> func (NodeHash) IsNode() bool { return false }
> ```

Should return `true`. Looks like the same bug is in Mongo.

---

*[graph/rethinkdb/quadstore.go, line 127 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVFx8EIlcbhR-2PN_U:-KQVFx8EIlcbhR-2PN_V:b-yal7gg) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L127)):*
> ```Go
> }
> 
> func createNewRethinkDBGraph(addr string, options graph.Options) (err error) {
> ```

You are not using `err = ` here. Drop name from the signature.

Also, this function should fail if database already exists.

---

*[graph/rethinkdb/quadstore.go, line 131 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVKCZD2jZ470-pbR6Z:-KQVKCZD2jZ470-pbR6_:b-vx70gi) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L131)):*
> ```Go
> 	if err != nil {
> 		return
> 	}
> ```

defer session.Close()

---

*[graph/rethinkdb/quadstore.go, line 142 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVGL9WACTe68pP3Uo6:-KQVGL9WACTe68pP3Uo7:b-32uy9s) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L142)):*
> ```Go
> 
> 	session, err = openSession(addr, dbName)
> 	if err != nil {
> ```

Strange condition. Drop names from signature and just return here.

---

*[graph/rethinkdb/quadstore.go, line 149 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVGbyYAO4FLZn3k_WM:-KQVGbyZgCF9hK5dweZY:b-45dnfw) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L149)):*
> ```Go
> 
> func newQuadStore(addr string, options graph.Options) (qs graph.QuadStore, err error) {
> 	s := QuadStore{}
> ```

Move to the end with inline struct initialization in return.

---

*[graph/rethinkdb/quadstore.go, line 156 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVK3YLW29ODJFEL8zP:-KQVK3YLW29ODJFEL8zQ:b-yhdtp3) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L156)):*
> ```Go
> 
> 	if err = ensureIndexes(session); err != nil {
> 		return
> ```

Close session

---

*[graph/rethinkdb/quadstore.go, line 177 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV3g1-Qvej5tSm17BG:-KQV3g1-Qvej5tSm17BH:bfexhst) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L177)):*
> ```Go
> 	hash += hashOf(t.Object)
> 	hash += hashOf(t.Label)
> 	h := sha1.New()
> ```

Hmm, interesting. But maybe we can omit rehashing?

---

*[graph/rethinkdb/quadstore.go, line 188 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVKtfRx4pY_Pc5jZs0:-KQVKtfRx4pY_Pc5jZs1:b69jxu2) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L188)):*
> ```Go
> 		ID:   string(node.(NodeHash)),
> 		Name: toRethinkDBValue(name),
> 		Size: inc,
> ```

Looks like a bug. This should increment current value, not overwrite it. We need a test for it in `graphtest` package.

---

*[graph/rethinkdb/quadstore.go, line 317 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVLscXXmzkDIo9B9qQ:-KQVLscXXmzkDIo9B9qR:b7daelx) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L317)):*
> ```Go
> 	switch d := v.(type) {
> 	case quad.Raw:
> 		return string(d) // compatibility
> ```

This was for compatibility. It should use []byte(d) in fact, if Rethink supports it.

---

*[graph/rethinkdb/quadstore.go, line 351 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVMHt3VJS0swBFhgai:-KQVMHt3VJS0swBFhgaj:b-oj1mwp) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L351)):*
> ```Go
> 	}
> 	switch d := v.(type) {
> 	case string:
> ```

Same here. If []byte, then...

---

*[graph/rethinkdb/quadstore.go, line 445 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVHOPj5QX6GuMcUFQT:-KQVHOPkNDViAD4eZU_D:bky3shw) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L445)):*
> ```Go
> 	switch err {
> 	case nil: // do nothing
> 	case gorethink.ErrEmptyResult: // do nothing
> ```

Return here?

---

*[graph/rethinkdb/quadstore_test.go, line 45 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV6OfAP6Yjf9wBT2t_:-KQV6OfAP6Yjf9wBT2ta:b-zhgknz) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore_test.go#L45)):*
> ```Go
> 		SkipSizeCheckAfterDelete: true,
> 		SkipNodeDelAfterQuadDel:  true,
> 	})
> ```

Is this config intentional, or it's just a Mongo legacy?

---

*[graph/rethinkdb/rdb_utils.go, line 18 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVFIo4wNC8F3tCS20_:-KQVFIo5cPt5tlraii5x:bdjpe3z) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L18)):*
> ```Go
> 			Database: database,
> 			MaxIdle:  10,
> 			MaxOpen:  10,
> ```

Maybe leave defaults, so driver can set what works best. Or this is better according to our benchmarks?

---

*[graph/rethinkdb/rdb_utils.go, line 23 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV8h3Oj3pDoOtHtBVd:-KQV8h3Oj3pDoOtHtBVe:b6izsw0) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L23)):*
> ```Go
> 			break
> 		} else {
> 			time.Sleep(time.Second * 2)
> ```

This case should fail fast depending on the error type. And why it's necessary? Docker test runner should use it, but for other cases it's not an option.

---

*[graph/rethinkdb/rdb_utils.go, line 34 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVIO0Y8Qa6WKQTlAYf:-KQVIO0Y8Qa6WKQTlAYg:brvrdxj) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L34)):*
> ```Go
> 	exists, err := databaseExists(database, c)
> 	if err != nil {
> 		return
> ```

Close session here

---

*[graph/rethinkdb/rdb_utils.go, line 40 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVIRLKjoTABXI4gWPE:-KQVIRLKjoTABXI4gWPF:b-yhdtp3) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L40)):*
> ```Go
> 		err = gorethink.DBCreate(database).Exec(c)
> 		if err != nil {
> 			return
> ```

Close session

---

*[graph/rethinkdb/rdb_utils.go, line 51 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQV6jCgrqOcMVjAdSsg:-KQV6jCgrqOcMVjAdSsh:b-v6j54a) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L51)):*
> ```Go
> // databaseExists checks if a database exists
> func databaseExists(name string, s *gorethink.Session) (exists bool, err error) {
> 	exists = false
> ```

False already

---

*[graph/rethinkdb/rdb_utils.go, line 95 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVEaoGrWmnzFi7Or3b:-KQVEaoGrWmnzFi7Or3c:b-y3rtx0) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L95)):*
> ```Go
> func ensureTable(name string, s *gorethink.Session) (err error) {
> 	var exists bool
> 	if exists, err = tableExists(name, s); err != nil || exists {
> ```

Let us track less variables: replace with `:=` and `return err`

---

*[graph/rethinkdb/rdb_utils.go, line 99 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVEkMJ5SiTNDvZl52-:-KQVEkMJ5SiTNDvZl520:bfie1qt) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/rdb_utils.go#L99)):*
> ```Go
> 	}
> 
> 	err = gorethink.TableCreate(name).Exec(s)
> ```

Same here - just return. And the same changes for other similar functions below.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KQVJ5dhUicOUAof6GQU:b-y7ponn)*

dennwc:
Reviewed 8 of 8 files at r2.

---

*[graph/rethinkdb/quadstore.go, line 156 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQVK3YLW29ODJFEL8zP:-KQXz4LxGLJ657Hp9GWj:b-kdbync) ([raw file](https://github.com/cayleygraph/cayley/blob/3d8d3f121d8cb39cfc1228c0dfc6d3ec40e7e0f7/graph/rethinkdb/quadstore.go#L156)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary>
> Close session
</details>
Session is still not closed here

---

*[graph/rethinkdb/quadstore.go, line 199 \[r2\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQXyha_uxLoiStKH8Qy:-KQXxukLWzybsQxKOY8_:b6md077) ([raw file](https://github.com/cayleygraph/cayley/blob/572bf0e48bd2e82528d8ac743e3c241fa3c5c216/graph/rethinkdb/quadstore.go#L199)):*
> ```Go
> func (qs *QuadStore) getIDForQuad(t quad.Quad) string {
> 	h := sha1.New()
> 	h.Write([]byte(quad.HashOf(t.Subject)))
> ```

Hashes twice, but it's more effective at least. And you can omit type conversion - HashOf returns []byte already.

---

*[graph/rethinkdb/quadstore_test.go, line 34 \[r2\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQXw4ySsy0jX_0SEfTr:-KQXw4ySsy0jX_0SEfTs:b1ac9ma) ([raw file](https://github.com/cayleygraph/cayley/blob/572bf0e48bd2e82528d8ac743e3c241fa3c5c216/graph/rethinkdb/quadstore_test.go#L34)):*
> ```Go
> 
> 	addr, closer := dock.Run(t, conf)
> 	addr = "localhost" + ":28015"
> ```

Any reasons for forcing it to an exposed host port? Tests can't be run in parallel with this option enabled.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468)*

dennwc:
Base64 is OK for hashes. The alternative is only hex and it will be larger anyway.

---

Reviewed 2 of 2 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[graph/rethinkdb/iterator.go, line 75 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQe2C9H69PGW2qqgt7Y:-KQe2C9H69PGW2qqgt7Z:b5jqrzr) ([raw file](https://github.com/cayleygraph/cayley/blob/cd4957e26aacbc47b2d2b510485d06b888d4d34e/graph/rethinkdb/iterator.go#L75)):*
> ```Go
> 		iter:       nil,
> 		size:       -1,
> 		hash:       NodeHash{},
> ```

Safe to omit. Same for iter.

---

*[graph/rethinkdb/quadstore.go, line 443 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQe6LZn0kEoQTTkWz9C:-KQe6LZn0kEoQTTkWz9D:bumv433) ([raw file](https://github.com/cayleygraph/cayley/blob/cd4957e26aacbc47b2d2b510485d06b888d4d34e/graph/rethinkdb/quadstore.go#L443)):*
> ```Go
> 	case quad.Time:
> 		n.Type = dbTime
> 		n.TimeValue = time.Time(d).UTC()
> ```

Test fails with UTC and passes without (with all UTC calls removed)

---

*[graph/rethinkdb/quadstore_test.go, line 15 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQe78ZCTGMuMRMyLMhe:-KQe78ZCTGMuMRMyLMhf:blnbmzv) ([raw file](https://github.com/cayleygraph/cayley/blob/cd4957e26aacbc47b2d2b510485d06b888d4d34e/graph/rethinkdb/quadstore_test.go#L15)):*
> ```Go
> 
> func makeRethinkDB(t testing.TB) (graph.QuadStore, graph.Options, func()) {
> 	clog.SetV(5)
> ```

Please remove this after everything else is done.

---

*[graph/rethinkdb/quadstore_test.go, line 37 \[r4\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQe5InSRpbh233pntZD:-KQe5InSRpbh233pntZE:bsukq1k) ([raw file](https://github.com/cayleygraph/cayley/blob/cd4957e26aacbc47b2d2b510485d06b888d4d34e/graph/rethinkdb/quadstore_test.go#L37)):*
> ```Go
> 	addr, closer := dock.Run(t, conf)
> 	//addr += ":28015"
> 	addr = "localhost:28015" // TODO: Use docker addr (see TODO above)
> ```

Double checked - test runs fine on my machine even without exposed ports. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KQe8GcHMQYe62_tiy_e:b-esbop3)*

dennwc:
All core tests are green, now it works as expected.

I forgot to mention that it's also a good idea to include a database version number into a separate table. We will need this information later to be able to upgrade in case of structural changes.

The next part is the performance. I tried to run our integration tests (/integration) and was bored to wait until it loads the initial dataset. The import performance is really bad right now. But the solution is simple - batch everything. I modified the code to add all log entries in one insert and write performance jumped from 20 writes/sec to 500 writes/sec. Try to batch all modifications in ApplyDeltas, if possible.

---

Reviewed 3 of 4 files at r5, 1 of 2 files at r6, 2 of 2 files at r7.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KQgoNsDQynJ2rpKDOtD:bbrtm3a)*

dennwc:
Looks good! Few small comments here.

Also, I tried to run integration tests today, and it took 3 hours to get to the 4th test only. So the next step is to optimize queries or minimize roundtrips if possible. Also I noticed that RethinkDB docs says to explicitly specify what index to use. If it's really the case and it's not used automatically, this may be the cause.

---

Reviewed 1 of 2 files at r8, 1 of 1 files at r9.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[graph/rethinkdb/quadstore.go, line 29 \[r8\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQkjw6SR4kdEc3EeZAP:-KQkjw6SR4kdEc3EeZAQ:b23ol82) ([raw file](https://github.com/cayleygraph/cayley/blob/7badbfb8dd38cbddc99b88d4a423d0191f885894/graph/rethinkdb/quadstore.go#L29)):*
> ```Go
> 	quadTableName    = "quads"
> 	logTableName     = "log"
> 	versionTableName = "version"
> ```

Maybe "meta" or "metadata", since it's a global KV for a database in general.

---

*[graph/rethinkdb/quadstore.go, line 212 \[r8\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQkkIkN5kDmTkdXWQY8:-KQkkIkN5kDmTkdXWQY9:b-7ju4bz) ([raw file](https://github.com/cayleygraph/cayley/blob/7badbfb8dd38cbddc99b88d4a423d0191f885894/graph/rethinkdb/quadstore.go#L212)):*
> ```Go
> 
> 	if dbVersion.Value != version && clog.V(2) {
> 		clog.Warningf("RethinkDB stored version: %d != implementation version: %d", dbVersion.Value, version)
> ```

This should always return an error. We don't know how database structure will change later, thus it's safer to tell the user that he must run a cayleyupgrade tool to convert the database.

---

*[graph/rethinkdb/quadstore.go, line 303 \[r8\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQklZJaDwu3NHUjb3wi:-KQklZJaDwu3NHUjb3wj:b-hxtk0q) ([raw file](https://github.com/cayleygraph/cayley/blob/7badbfb8dd38cbddc99b88d4a423d0191f885894/graph/rethinkdb/quadstore.go#L303)):*
> ```Go
> 		data, err := proto.MakeValue(v).Marshal()
> 		if err != nil {
> 			clog.Errorf("Failed to marshal value")
> ```

Print the error also

---

*[graph/rethinkdb/quadstore.go, line 315 \[r8\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQkmC6Qyd3jxRZf5zix:-KQkmC6Qyd3jxRZf5ziy:b-9b10ux) ([raw file](https://github.com/cayleygraph/cayley/blob/7badbfb8dd38cbddc99b88d4a423d0191f885894/graph/rethinkdb/quadstore.go#L315)):*
> ```Go
> 	var logEntries []LogEntry
> 	var quads []Quad
> 	var nodes []Node
> ```

Merge into one var block

---

*[graph/rethinkdb/quadstore.go, line 215 \[r9\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQkm_ajQwLcNLkF51H0:-KQkm_ajQwLcNLkF51H1:b-n1i17t) ([raw file](https://github.com/cayleygraph/cayley/blob/afa2a4db0683187dec17a5b0ab8cbe8b32845de0/graph/rethinkdb/quadstore.go#L215)):*
> ```Go
> 			clog.Infof("RethinkDB stored version: %d != implementation version: %d", dbVersion.Value, version)
> 		}
> 		if err = performUpgrade(dbVersion.Value, session); err != nil {
> ```

This may take a long time. We have a special tool for this, called cayleyupgrade. Check UpgradeFunc in RegisterQuadStore. And it's okay to omit it right now, since we don't have any new versions yet. Just return an error that the version is unknown.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KQkn9-mrhcALIcSnlC-:b3ldh1g)*

dennwc:
Does it helps to use the index? I mean, I was a bit surprised that DB will not use it automatically.

---

Reviewed 4 of 4 files at r10.

---

*[graph/rethinkdb/quadstore.go, line 26 \[r10\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQlGlo_nzsGAuCpKykH:-KQlGloakdWevlcYgg_J:b25hl8g) ([raw file](https://github.com/cayleygraph/cayley/blob/097707aa94a18b1bfa56bb6305323adefbdc20b6/graph/rethinkdb/quadstore.go#L26)):*
> ```Go
> 
> const (
> 	nodeTableName     = "graph_nodes"
> ```

Why to add a prefix? It will already be in "cayley" database. If you need it for some reason, it may be better to add an option for it (defaults to empty).

---

*[graph/rethinkdb/quadstore.go, line 37 \[r10\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQlGBzV4q4RLh2ci7j8:-KQlGBzV4q4RLh2ci7j9:b-kq3bn4) ([raw file](https://github.com/cayleygraph/cayley/blob/097707aa94a18b1bfa56bb6305323adefbdc20b6/graph/rethinkdb/quadstore.go#L37)):*
> ```Go
> 		NewFunc:           newQuadStore,
> 		NewForRequestFunc: nil,
> 		UpgradeFunc:       upgradeRethinkDBGraph,
> ```

No, this will confuse cayleyupgrade. Function is here, but returns nil without doing anything. Just remove the function and put nil.

---

*[graph/rethinkdb/quadstore.go, line 214 \[r10\]](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KQlFuJnprYNx44_b4tF:-KQlFuJoT2iGc3mxFsps:b-cis11m) ([raw file](https://github.com/cayleygraph/cayley/blob/097707aa94a18b1bfa56bb6305323adefbdc20b6/graph/rethinkdb/quadstore.go#L214)):*
> ```Go
> 		err = fmt.Errorf("RethinkDB version mismatch. DB: %d != impl.: %d", dbVersion.Value, version)
> 		clog.Errorf("%v", err)
> 		return
> ```

Don't forget to close the session ;)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KQlHODvn7eJ4qvJzk4K:btlgek1)*

dennwc:
Looks good! Let me check tests and I think we can merge it.

---

Reviewed 1 of 1 files at r11, 2 of 2 files at r12, 4 of 4 files at r13.

---

*[graph/rethinkdb/quadstore.go, line 115 at r10](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRbw9mSYM1GxNITT4VV:-KRbw9mTHSOIfH6Hd76r:bxgzj59) ([raw file](https://github.com/cayleygraph/cayley/blob/097707aa94a18b1bfa56bb6305323adefbdc20b6/graph/rethinkdb/quadstore.go#L115)):*
> ```Go
> 	Object    string   `json:"object"`
> 	Label     string   `json:"label"`
> 	Added     []string `json:"added"`
> ```

Without these fields LogEntry becomes nearly useless. But we are not using any functions for getting log in any of backends, so it might be ok for now.

---

*[graph/rethinkdb/quadstore_test.go, line 18 at r13](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRbygEoaN7cyKJ5utGp:-KRbygEoaN7cyKJ5utGq:b-r6ptex) ([raw file](https://github.com/cayleygraph/cayley/blob/9d4f76ea20fc4fbfee2a9fc00a8368c2b26f5004/graph/rethinkdb/quadstore_test.go#L18)):*
> ```Go
> 
> func init() {
> 	clog.SetV(3)
> ```

Please do not set this in tests - it will spam logs. Use it for debugging purposes only.

---

*[graph/rethinkdb/quadstore_test.go, line 39 at r13](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRbyC1PvNQkKBVOxl8C:-KRbyC1PvNQkKBVOxl8D:bt813be) ([raw file](https://github.com/cayleygraph/cayley/blob/9d4f76ea20fc4fbfee2a9fc00a8368c2b26f5004/graph/rethinkdb/quadstore_test.go#L39)):*
> ```Go
> 		addr = fmt.Sprintf("%s:%d", addr, 28015)
> 	} else {
> 		// If running test on Mac (Mac for Docker) we need to bind ports to host
> ```

This can be actually useful for other tests as well. Will add it to dock library later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KRbzTN8wNb75Xr47paz:b3p22r8)*

dennwc:
Looks very promising! Few more comments, especially on concurrency. And I will try to run benchmarks again with a new implementation :) 

---

Reviewed 7 of 7 files at r14, 1 of 1 files at r15.
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[graph/rethinkdb/indexed_linksto.go, line 58 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRk5KxkaqmYogfmg4LY:-KRk5KxkaqmYogfmg4LZ:b-clj8d8) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/indexed_linksto.go#L58)):*
> ```Go
> 
> 	// reversed
> 	if index, ok := dirLinkIndexMap[[2]quad.Direction{dir2, dir2}]; ok {
> ```

"dir2, dir1" probably?

---

*[graph/rethinkdb/quadstore.go, line 113 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRrItrPJTS1D_M-KJDF:-KRrItrQh5e0-DijJX67:b-gt5rbl) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/quadstore.go#L113)):*
> ```Go
> 
> var dirLinkIndexMap = map[[2]quad.Direction]string{
> 	{quad.Subject, quad.Subject}:     "subject_subject",
> ```

Why it has duplicated directions in index? Am I missing something?

---

*[graph/rethinkdb/quadstore.go, line 195 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRk7QAIEHY_z8VnXzXj:-KRk7QAIEHY_z8VnXzXk:bms7mqz) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/quadstore.go#L195)):*
> ```Go
> 		return
> 	}
> 	for _, dir := range []string{quad.Subject.String(), quad.Predicate.String(), quad.Object.String(), quad.Label.String()} {
> ```

Too long. Call String in the line below.

---

*[graph/rethinkdb/quadstore.go, line 504 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRk7nWBO78EfRAC8ahU:-KRk7nWBO78EfRAC8ahV:b-gfec9h) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/quadstore.go#L504)):*
> ```Go
> 	wg.Add(len(queries))
> 	for _, q := range queries {
> 		go func(query gorethink.Term) {
> ```

It's good for batch inserts, but it may cause troubles in other cases. Maybe add some kind of gate to limit these?

---

*[graph/rethinkdb/quadstore.go, line 510 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRk83QPjurt8iIoMiPi:-KRk83QPjurt8iIoMiPj:b-dwxips) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/quadstore.go#L510)):*
> ```Go
> 				errCh <- err
> 			}
> 			wg.Done()
> ```

Better to defer it at the beginning.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KRrJR_Yr_YE2v8ySZxM:b-61t429)*

dennwc:
Reviewed 2 of 2 files at r16.

---

*[graph/rethinkdb/quadstore.go, line 504 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRk7nWBO78EfRAC8ahU:-KRt6CEyfAcpYmoL1bfj:b-vh47z) ([raw file](https://github.com/cayleygraph/cayley/blob/b0c2d1d891ffef8410d77958f1a59f5897ef539d/graph/rethinkdb/quadstore.go#L504)):*
<details><summary><i>Previously, dennwc (Denys Smirnov) wrote…</i></summary>
> It's good for batch inserts, but it may cause troubles in other cases. Maybe add some kind of gate to limit these?
</details>
It's not really limiting. I can still send a set of large deltas that will force parallel mode and it will spawn tens of goroutines regardless of the max_connections option. I think we can do this way: if limit is 1 - run in linear mode, and if limit > 1 - setup a channel, run the same number of workers (equal to min of limit and len(queries) )  and let them take queries from the channel. In this case it would be clear how many queries will be executed per ApplyDelta call.

---

*[graph/rethinkdb/quadstore.go, line 307 at r16](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KRt7bjS6xrTE9sqRDBC:-KRt7bjTvhmh0EFbQQxP:b2a6tjf) ([raw file](https://github.com/cayleygraph/cayley/blob/9a87565c4d4fbc339f8b2b5d1ca2f80816483423/graph/rethinkdb/quadstore.go#L307)):*
> ```Go
> 
> 	maxConnections := 2
> 	if val, ok, err := options.IntKey("max_connections"); err == nil && ok {
> ```

Name overlaps with an option from dialRethinkDB. Is this intentional?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468)*

dennwc:
Sure. I'm thinking of ways to move parallel approach higher, because I also implemented it for PostgreSQL in the fork a while ago. It would be easier to maintain.

If optimization step is done, I'll check the performance on my machine also.

As I see, Rethink has a pretty flexible query language, so maybe there are some other things that we can do, like optimize one of these iterators: And/Or/Skip/Limit? Is it an easy task to do?

---

Reviewed 1 of 1 files at r17.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KRtFFYRAlZR_m7GtViB:b-ajs63e)*

dennwc:
@mikaelcabot Sorry, was a bit busy. Will review and test it in a few days.

dennwc:
Reviewed 2 of 2 files at r18, 1 of 2 files at r20.
Review status: 9 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[graph/rethinkdb/quadstore_test.go, line 97 at r20](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KSh7krSQqjYNnXwzza-:-KSh7krSQqjYNnXwzza0:b-87r8nn) ([raw file](https://github.com/cayleygraph/cayley/blob/af71ccdd08318f229a9e0597d701fbfe2f79bb95/graph/rethinkdb/quadstore_test.go#L97)):*
> ```Go
> }
> 
> func makeQuadSet() []quad.Quad {
> ```

The same function should be available in graphtest package. Or was it in one of recent commits?

---

*[graph/rethinkdb/quadstore_test.go, line 131 at r20](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KSh6l-ZGana8k-a20ZE:-KSh6l-_eFCCoCgI24CW:b-wqzsb8) ([raw file](https://github.com/cayleygraph/cayley/blob/af71ccdd08318f229a9e0597d701fbfe2f79bb95/graph/rethinkdb/quadstore_test.go#L131)):*
> ```Go
> 
> 	it := iterator.NewLimit(allIt, int64(limit))
> 	it.Optimize()
> ```

It will not optimize Limit for QuadStore. It's better to type-assert it to Rethink-specific iterator to be sure. Same for Skip.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468)*

dennwc:
Reviewed 1 of 2 files at r20, 2 of 2 files at r21.

---

*[graph/rethinkdb/quadstore_iterator_optimize.go, line 219 at r18](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KSlgej2IvabZEVhREQp:-KSlgej3DDHDheG9tmGN:bjvhvn5) ([raw file](https://github.com/cayleygraph/cayley/blob/4b8ac21a7dd0461c96fa49a60090c1bfe8ddf9ff/graph/rethinkdb/quadstore_iterator_optimize.go#L219)):*
> ```Go
> 		return it, false
> 	}
> 	primaryIt.query = primaryIt.query.Skip(it.Size())
> ```

What is the problem with Skip? Why it's removed now? Is there any problems with it?

---

*[graph/rethinkdb/quadstore_test.go, line 131 at r20](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-KSh6l-ZGana8k-a20ZE:-KSlgEr-2Bf2HbEKyQhc:b-dng0ar) ([raw file](https://github.com/cayleygraph/cayley/blob/af71ccdd08318f229a9e0597d701fbfe2f79bb95/graph/rethinkdb/quadstore_test.go#L131)):*
> qs.OptimizeIterator(it)
It will return another iterator as a result, check the signature. You should test returned iterator, not an old one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468)*

dennwc:
Looks good overall. Query performance was 2 times worse comparing to Mongo, but this more likely caused by absence of caches. And I highly recommend to optimize And iterator at least for links.

Also we have some bad news:
https://rethinkdb.com/blog/rethinkdb-shutdown/

---

Reviewed 1 of 1 files at r22.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/468#-:-KTTHtUdilaYBqXb9C0l:b7egd4l)*

dennwc:
Great! I like your attitude :)

As a note, I have a version with CockroachDB support almost ready for testing.

So the situation with this PR is a bit controversial - we can merge it, but without Rethink being developed we might soon deprecate it. Let's see what happens next.

dennwc:
Sure, will file a PR soon

PR #470, opened 2016-08-31
Implement Recursive Iterator

barakmich:
Rebased and updated (now with types) my old `transitive_closure` branch. The notion here is that, in the path library:

```
g.V(start).FollowRecursive(g.M().Out("parent"))
```

Will follow the `parent` predicate (or general morphism) out to the set of reachable (within a maximum) nodes generated by repeating that query. There's the option for adding a special tag which will tell you, numerically, at what depth (eg, parent: 1, grandparent: 2, great-grandparent: 3) the node was first visited. Loops are _not_ followed; their depth is their minimum distance.

Effectively, it goes from an input set of nodes (left-hand side of the operation) to a set of nodes recursively reachable by the morphism (on the right hand side). 

Adding it to Gizmo is a followup.

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -265,6 +265,7 @@ var (
 		"unique",
 		"limit",
 		"skip",
+		"recursive",
 	}
 )
 
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -240,6 +240,7 @@ const (
 	Unique
 	Limit
 	Skip
+	Recursive
 )
 
 var (
--- a/graph/iterator/and_iterator_optimize_test.go
+++ b/graph/iterator/and_iterator_optimize_test.go
@@ -26,7 +26,7 @@ import (
 )
 
 func TestIteratorPromotion(t *testing.T) {
-	qs := &store{
+	qs := &oldstore{
 		data: []string{},
 		iter: NewFixed(Identity),
 	}
@@ -55,7 +55,7 @@ func TestIteratorPromotion(t *testing.T) {
 }
 
 func TestNullIteratorAnd(t *testing.T) {
-	qs := &store{
+	qs := &oldstore{
 		data: []string{},
 		iter: NewFixed(Identity),
 	}
@@ -74,7 +74,7 @@ func TestNullIteratorAnd(t *testing.T) {
 }
 
 func TestAllPromotion(t *testing.T) {
-	qs := &store{
+	qs := &oldstore{
 		data: []string{},
 		iter: NewFixed(Identity),
 	}
@@ -106,7 +106,7 @@ func TestAllPromotion(t *testing.T) {
 }
 
 func TestReorderWithTag(t *testing.T) {
-	qs := &store{
+	qs := &oldstore{
 		data: []string{},
 		iter: NewFixed(Identity),
 	}
@@ -145,7 +145,7 @@ func TestReorderWithTag(t *testing.T) {
 }
 
 func TestAndStatistics(t *testing.T) {
-	qs := &store{
+	qs := &oldstore{
 		data: []string{},
 		iter: NewFixed(Identity),
 	}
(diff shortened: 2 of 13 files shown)


dennwc:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[graph/iterator/recursive_iterator.go, line 125 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY6lrmylwT34nXHQBu:-KQY6lrmylwT34nXHQBv:bm4m0yi) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/graph/iterator/recursive_iterator.go#L125)):*
> ```Go
> 			it.depthCache = append(it.depthCache, it.subIt.Result())
> 			tags := make(map[string]graph.Value)
> 			it.subIt.TagResults(tags)
> ```

This step can be omitted in most cases, but we have no way to know if user will can TagResults on iterator later, correct?

---

*[graph/iterator/recursive_iterator.go, line 156 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY5_KTs5Ln-a1MfdA8:-KQY5_KTs5Ln-a1MfdA9:blyhri7) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/graph/iterator/recursive_iterator.go#L156)):*
> ```Go
> 			continue
> 		}
> 		it.seen[val] = seenAt{
> ```

This will panic on Bolt, because it uses non-comparable tokens. I made a `graph.ToKey` helper for this case. Same for pathMap.

---

*[graph/iterator/recursive_iterator.go, line 241 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY4zcMeCWT_Zo4IlxR:-KQY4zcMeCWT_Zo4IlxS:b-e7tmf7) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/graph/iterator/recursive_iterator.go#L241)):*
> ```Go
> 
> func (it *Recursive) Size() (int64, bool) {
> 	return it.Stats().Size, false
> ```

Maybe it's better to inline size calculations from Stats and use this function in Stats itself.

---

*[graph/path/path.go, line 309 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY09Oe6LMx-vGaLeHz:-KQY09Of66WbRS11XkPy:bozly6h) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/graph/path/path.go#L309)):*
> ```Go
> //
> // This is a very expensive operation in practice. Be sure to use it wisely.
> ```

Empty line prevents docs on method to be seen.

---

*[graph/path/path.go, line 313 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY4JfHPSzaCoMxV5qO:-KQY4JfIFC3zpNB6d5Kr:bhtlmyz) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/graph/path/path.go#L313)):*
> ```Go
> 	var path *Path
> 	switch v := via.(type) {
> 	case string:
> ```

What about quad.Value?

---

*[quad/value.go, line 15 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/470#-KQY1cHBSRQYcjfu7Ezf:-KQY1cHCk3HR_umarnPP:b-mx9wgd) ([raw file](https://github.com/cayleygraph/cayley/blob/824ff3e71ab5aee74e35ca814582e148ca459f5f/quad/value.go#L15)):*
> ```Go
> type Value interface {
> 	String() string
> 	IsNode() bool
> ```

This is done for any value to be compatible with `graph.Value`. But what kind of iterator can return or use these values? Is it only for testing with Fixed?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/470)*

dennwc:
Closing in favor of #547

PR #472, opened 2016-09-03
Etcd backend

dennwc:
Implementation of QuadStore on top of etcd v3.

Still needs some optimization.

### DIFF
--- a/cmd/cayley/cayley.go
+++ b/cmd/cayley/cayley.go
@@ -35,6 +35,7 @@ import (
 
 	// Load all supported backends.
 	_ "github.com/cayleygraph/cayley/graph/bolt"
+	_ "github.com/cayleygraph/cayley/graph/etcd"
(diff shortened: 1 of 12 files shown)


barakmich:
Closing for age, but also for the general notion **at this time**.

etcd --and i used to deal with it heavily!-- is primarily a _lockserver_ that appears as a key value store. Keys are much more atomic than evolving graphs. It's meant for metadata over, simply, data.

Will the world one day need a graph abstraction atop a lockserver? The Shadow knows. Until people can use a lockserver correctly, this is theoretical.

CockroachDB, while still early, is much more in the intended vein; strong consistency, but meant for large amounts of data. And, incidentally, based on the same Raft implementation. And we have that now, through their SQL interface. If we had a first-class interface to Cockroach's underlying database space, that'd be kickass.

So those options obviate etcd as an option. Closing, but with a few recourses open for comment.

PR #473, opened 2016-09-05
Add HasRegex and HasComparison paths

guycook:
This PR adds two new path traversals, HasRegex and HasComparison. In each case the caller provides an edge to look ahead, then the path is traversed if the found predicate passes either a regex match or numeric comparison respectively.

### DIFF
--- a/graph/bolt/all_iterator.go
+++ b/graph/bolt/all_iterator.go
@@ -96,12 +96,11 @@ func (it *AllIterator) Next() bool {
 			b := tx.Bucket(it.bucket)
 			cur := b.Cursor()
 			if last == nil {
-				k, _ := cur.First()
-				var out []byte
-				out = make([]byte, len(k))
-				copy(out, k)
-				it.buffer = append(it.buffer, out)
-				i++
+				k, v := cur.First()
+				if it.nodes || isLiveValue(v) {
+					it.buffer = append(it.buffer, clone(k))
+					i++
+				}
 			} else {
 				k, _ := cur.Seek(last)
 				if !bytes.Equal(k, last) {
@@ -109,15 +108,15 @@ func (it *AllIterator) Next() bool {
 				}
 			}
 			for i < bufferSize {
-				k, _ := cur.Next()
+				k, v := cur.Next()
 				if k == nil {
 					it.buffer = append(it.buffer, k)
 					break
 				}
-				var out []byte
-				out = make([]byte, len(k))
-				copy(out, k)
-				it.buffer = append(it.buffer, out)
+				if !it.nodes && !isLiveValue(v) {
+					continue
+				}
+				it.buffer = append(it.buffer, clone(k))
 				i++
 			}
 			return nil
--- a/graph/bolt/iterator.go
+++ b/graph/bolt/iterator.go
@@ -24,7 +24,6 @@ import (
 
 	"github.com/cayleygraph/cayley/graph"
 	"github.com/cayleygraph/cayley/graph/iterator"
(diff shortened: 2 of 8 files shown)


dennwc:
Although the idea of adding regexp iterator is great, I have to reject this pull request in its current state. It's made for a very limited use case and will not work for large sets at all. Also, it ignores the current implementation of comparison iterator for no good reason.

I will not close the PR, in case you want to update it with all the changes mentioned below.

---

Reviewed 3 of 3 files at r1.

---

*[graph/path/morphism_apply_functions.go, line 109 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/473#-KQyj1DkGlDDjeBkLjgT:-KQyj1DkGlDDjeBkLjgU:b-aybo4x) ([raw file](https://github.com/cayleygraph/cayley/blob/d8f965fce42aaeba188ab71b47b9bc438d12f7fb/graph/path/morphism_apply_functions.go#L109)):*
> ```Go
> 
> 			ends := func() graph.Iterator {
> 				re := regexp.MustCompile(pattern)
> ```

Doing regexp compilation on each iterator build is a bad idea. It should be in hasRegexMorphism function, because pattern is constant there.

---

*[graph/path/morphism_apply_functions.go, line 115 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/473#-KQyierubvgaPvAa6f8k:-KQyierveFZoHpahKSmC:b-8hneha) ([raw file](https://github.com/cayleygraph/cayley/blob/d8f965fce42aaeba188ab71b47b9bc438d12f7fb/graph/path/morphism_apply_functions.go#L115)):*
> ```Go
> 				// pass the regex.
> 				fixed := qs.FixedIterator()
> 				for out.Next() {
> ```

This is a bad. You may end up materializing all the database in memory. RegExp should be a separate iterator type and filter things on the fly.

---

*[graph/path/morphism_apply_functions.go, line 150 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/473#-KQyk0-KZ1hz62o1674y:-KQyk0-KZ1hz62o1674z:b-7btimf) ([raw file](https://github.com/cayleygraph/cayley/blob/d8f965fce42aaeba188ab71b47b9bc438d12f7fb/graph/path/morphism_apply_functions.go#L150)):*
> ```Go
> 			ends := func() graph.Iterator {
> 				var compare func(float64, float64) bool
> 				switch operator {
> ```

You are reinventing our Comparison iterator. This was already done in a more efficient way for all types that we support.

---

*[graph/path/path.go, line 350 \[r1\]](https://reviewable.io:443/reviews/cayleygraph/cayley/473#-KQyjqCPRnkwyoStnZfJ:-KQyjqCPRnkwyoStnZfK:bydi4t) ([raw file](https://github.com/cayleygraph/cayley/blob/d8f965fce42aaeba188ab71b47b9bc438d12f7fb/graph/path/path.go#L350)):*
> ```Go
> // HasComparison limits the paths to be ones where the current nodes have some linkage
> // to a node satisfying a numeric comparison.
> func (p *Path) HasComparison(via interface{}, operator string, number float64) *Path {
> ```

Why only numbers? What about strings, time, ...?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/473#-:-KQykQaBzC48V4Nx98bO:b-kc2nah)*

guycook:
Thanks for looking at that dennwc. I'm definitely going to try and fix this up as regex is quite an important feature for a project I'm working on.

guycook:
Alternate implementation provided as pr #474

PR #474, opened 2016-09-12
Implement regex filter as iterator

guycook:
Provides a new filtering path `.Regex`. The implementation largely mirrors that of `.Filter`

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -240,6 +240,7 @@ const (
 	Unique
 	Limit
 	Skip
+	Regex
 )
 
 var (
@@ -264,6 +265,7 @@ var (
 		"unique",
 		"limit",
 		"skip",
+		"regex",
 	}
 )
 
--- a/graph/iterator/regex_iterator.go
+++ b/graph/iterator/regex_iterator.go
@@ -14,11 +14,6 @@
 
 package iterator
 
-// "Regex" is a unary operator -- a filter across the values in the relevant
-// subiterator. It works similarly to gremlin's filter{it.matches('exp')},
-// reducing the iterator set to values whose string representation passes a
-// regular expression test.
-
 import (
 	"regexp"
 
@@ -26,6 +21,10 @@ import (
 	"github.com/cayleygraph/cayley/quad"
 )
 
+// Regex is a unary operator -- a filter across the values in the relevant
+// subiterator. It works similarly to gremlin's filter{it.matches('exp')},
+// reducing the iterator set to values whose string representation passes a
+// regular expression test.
 type Regex struct {
 	uid    uint64
 	tags   graph.Tagger
@@ -70,6 +69,8 @@ func (it *Regex) Close() error {
 
 func (it *Regex) Reset() {
 	it.subIt.Reset()
+	it.err = nil
+	it.result = nil
 }
 
 func (it *Regex) Tagger() *graph.Tagger {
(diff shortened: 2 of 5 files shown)


dennwc:
Well done! Looks much better now. Few minor changes (which I should apply to Comparison also) and I'll merge it.

Also please add Regex to Gremlin, or as an alternative, I can add it after the merge.

---

Reviewed 5 of 5 files at r1.

---

*[graph/iterator/regex_iterator.go, line 17 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-KRSt7aDlLhgVMoct3n-:-KRSt7aDlLhgVMoct3n0:bwtegrb) ([raw file](https://github.com/cayleygraph/cayley/blob/ace4309621be2fb564ac24641796e14e00356750/graph/iterator/regex_iterator.go#L17)):*
> ```Go
> package iterator
> 
> // "Regex" is a unary operator -- a filter across the values in the relevant
> ```

Put this comment on top of Regex struct definition, so it can appear in godoc. Also, remove `"` around Regex to match Go convention for comments.

---

*[graph/iterator/regex_iterator.go, line 72 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-KRSsTB5QSFyVNnv_yjC:-KRSqi77-AwjAXhRsRUB:birirkg) ([raw file](https://github.com/cayleygraph/cayley/blob/ace4309621be2fb564ac24641796e14e00356750/graph/iterator/regex_iterator.go#L72)):*
> ```Go
> 
> func (it *Regex) Reset() {
> 	it.subIt.Reset()
> ```

Also set a it.result and it.err to nil

---

*[graph/iterator/regex_iterator.go, line 102 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-KRStXe3HrD6TeCbq7rb:-KRStXe3HrD6TeCbq7rc:b-ty2nyh) ([raw file](https://github.com/cayleygraph/cayley/blob/ace4309621be2fb564ac24641796e14e00356750/graph/iterator/regex_iterator.go#L102)):*
> ```Go
> 
> func (it *Regex) Result() graph.Value {
> 	return it.result
> ```

We can potentially drop it.err and it.result and return them from sub iterator directly.

---

*[graph/iterator/regex_iterator.go, line 180 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-KRSuvCJKZ9FSnM7VXTg:-KRSrYsr6cL6H9quCbS5:b-cnt951) ([raw file](https://github.com/cayleygraph/cayley/blob/ace4309621be2fb564ac24641796e14e00356750/graph/iterator/regex_iterator.go#L180)):*
> ```Go
> 
> func (it *Regex) Size() (int64, bool) {
> 	return 0, true
> ```

Returning `true` here means that result is exact. Optimizer may think that this iterator is empty and remove it from the tree. I should fix this for Comparison also.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-:-KRStn29zzcbBupusFHU:bd0uquf)*

guycook:
I went for the 'set err/result to nil' approach instead of returning directly from subiterator. I feel like I would need to do some more comprehensive tests for that case.

If you could add the Gremlin API that would be great - I'm not too familiar with the code there and unfortunately quite time constrained at the moment.

dennwc:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/474#-:-KRdt7FfB3dl6Ep2y_pz:bnfp4nl)*

Issue #353, opened 2015-12-05
query: Schema/RDFS


dennwc:
I've been working on a schema layer for Cayley recently to make querying/writing structured data easier. I want to discuss it's design, if someone interested in it as I do. It can be really useful, because you still can do sophisticated graph queries and then query structured data for results.

For example, lets consider this data set. It describes a `Thing` class with two properties.

```
<http://schema.org/Thing> <http://www.w3.org/2000/01/rdf-schema#label> "Thing" .
<http://schema.org/Thing> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2000/01/rdf-schema#Class> .
<http://schema.org/Thing> <http://www.w3.org/2000/01/rdf-schema#comment> "The most generic type of item." .

<http://schema.org/url> <http://schema.org/domainIncludes> <http://schema.org/Thing> .
<http://schema.org/url> <http://www.w3.org/2000/01/rdf-schema#label> "url" .
<http://schema.org/url> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/1999/02/22-rdf-syntax-ns#Property> .
<http://schema.org/url> <http://schema.org/rangeIncludes> <http://schema.org/URL> .
<http://schema.org/url> <http://www.w3.org/2000/01/rdf-schema#comment> "URL of the item." .

<http://schema.org/name> <http://schema.org/domainIncludes> <http://schema.org/Thing> .
<http://schema.org/name> <http://www.w3.org/2000/01/rdf-schema#label> "name" .
<http://schema.org/name> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/1999/02/22-rdf-syntax-ns#Property> .
<http://schema.org/name> <http://schema.org/rangeIncludes> <http://schema.org/Text> .
<http://schema.org/name> <http://www.w3.org/2000/01/rdf-schema#comment> "The name of the item." .
```

I came up with a few rules for mapping quads to fields:
- `"."` tag means "save current node to this field"
- `"pred"` (`"pred > *"` in full form) means "save object(s) with this predicate"; maps to `path.Save`
- `"pred < *"` is a reverse of previous action; maps to `SaveReverse`
- `"pred > obj1,obj2"` means "filter nodes that have a specific object on predicate"; maps to `Has`
- `"pred < obj1,obj2"` is a reverse of previous action; maps to `HasReverse`
[...]

PR #475, opened 2016-09-14
Schema lib (aka ORM, aka Object Persistence Layer)

dennwc:
A backport of idea described in [#353](https://github.com/cayleygraph/cayley/issues/353).

Library does two things:
- Converts any Go object (with annotations) to a set of quads.
- Builds and execute queries to get the same objects from the database. 

In short, it can convert struct:

``` Go
type Item struct {
    rdfType struct{} `quad:"rdf:type > some:item"`
    ID      quad.IRI `quad:"@id"`
    Name    string   `quad:"name"`
    Spec    []string   `quad:"spec,optional"`
}
```

Into quads:

``` nquad
<some_id> <rdf:type> <some:item>
<some_id> <name> "Test name"
<some_id> <spec> "spec1"
<some_id> <spec> "spec2"
```

And query it back from the database given the same `<some_id>` and a Go struct type.

More details can be found in [original proposal](https://github.com/cayleygraph/cayley/issues/353).

As a side effect, library implements an RDF vocabulary registry (as a separate package) to expand and compact IRIs.


abourget:
Couldn't:

```
rdfType struct{} `quad:"rdf:type > some:item"`
```

be simplified ? or do we need that `rdf:type > ` notation in there ?

dennwc:
@abourget Could be simplified to:
```Go
_ struct{} `quad:"@type > some:item"`
```
This really makes a difference, right? :grin:

Actually, with the lastest commit, Go type can be bound to IRI like this:
```Go
func init(){
   schema.RegisterType(quad.IRI("some:item"), Item{})
}
```
It's essentially the same for the system - it will then generate an additional type triple each time object of type `Item` is saved and enforce the same constraint when object is loaded.
I think it would be much easier to work with then previous mechanism with struct tags. What do you think?

It worth to mention that lib can save/load structs even without a type. In this case it matches struct fields only (if they were not defined as optional).

abourget:
yeah, like Gob registrations.. easier for newcomers.

faddat:
I'm building a cayley that has this in it now.  Any idea when this might get into mainline?

dennwc:
I consider it done. Feel free to comment and let's merge it!

robertmeta:
Let me try this out, \o/

PR #489, opened 2016-11-06
Support SQL flavors: Mysql, CockroachDB

dennwc:
Adds `flavor` option to sql backend. Allowed values: `postgres` (default), `mysql`, `cockroach`.

**TODO**

- [ ] check integration tests

- [ ] update docs with examples of each configuration

- [ ] check [transaction retries](https://www.cockroachlabs.com/docs/transactions.html#transaction-retries) for CockroachDB

- [ ] test large queries in CockroachDB and maybe disable JOINs

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -48,8 +48,11 @@ import (
 
 const format = "cquad"
 
-var backend = flag.String("backend", "memstore", "Which backend to test. Loads test data to /tmp if not present.")
-var backendPath = flag.String("backend_path", "", "Path to the chosen backend. Will have sane testing defaults if not specified")
+var (
+	backend     = flag.String("backend", "memstore", "Which backend to test. Loads test data to /tmp if not present.")
+	backendPath = flag.String("backend_path", "", "Path to the chosen backend. Will have sane testing defaults if not specified")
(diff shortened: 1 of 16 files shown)


irfansharif:
[cockroachdb/cockroach/pull/10618](https://github.com/cockroachdb/cockroach/pull/10618) just went in yesterday. to your point above to 'maybe disable JOINS' for now if your JOIN queries are of the form `SELECT ... JOIN ... USING(...)` it'll be optimal (using the O(n) hash join algorithm). Equality constraints in the `ON` predicate (i.e. `ON a.x = b.x`) for now default to the non-optimal nested loop implementation but this is what we're working on now, should be performant as well.

integration/integration_test.go
```diff
@@ -462,6 +465,9 @@ func prepare(t testing.TB) {
 		remote = true
 	case "sql":
 		cfg.DatabasePath = "postgres://localhost/cayley_test"
```
barakmich: Can we infer flavor from the protocol part of the URL, if not explicitly stated?

dennwc: CockroachDB is the same as postgres and mysql has no protocol part in URL. But we can define them ourselves, so they are more consistent.
I'm okay with flavor option, since I can copy address directly from other SQL packages for Go. But not a big deal.

PR #492, opened 2016-11-13
New CLI

dennwc:
Switch to Cobra CLI library.

- More predictable flags handling.
- Shortcuts for frequently used flags.
- Better help for individual commands.
- `cayleyupgrade` is now included into main binary.
- Add command to convert quad files between formats without loading them (batch processing).
- Config format changed, but old one is still supported.
- Updated docs.


barakmich:
LGTM when comment resolved

dennwc:
@barakmich Thought about it, but it seems that we will have more remote backends than a local ones soon. I'm not opposed to `path` but, it appears more logical to me to specify address `/path/to/db` (`file:///path/to/db`?), then path `127.0.0.1:2379`, for example. What do you think? 

Also, it's still a breaking change for anyone using CLI, thus we might want to merge few other things, cut a release and then merge this one. Does it makes sense?

barakmich:
Hm. Can we combine them? Make `-path /foo/bar` an alias into -address `file:///foo/bar` and let `address` be canonical?

dennwc:
Probably yes, but we'll need to remember to call some helper each time we access `address`. Maybe just make them aliases, and that's all? Backend will check if it's correct one or not anyway. Right now I made `store.path` and `store.address` aliases (`address` is primary), but flag is still called `dbpath`, which probably should be, `address` with an alias `path` (if Cobra supports itl).

PR #493, opened 2016-11-16
Moving documentation to discourse for now

robertmeta:
I am looking to move all the documentation to discourse for us to get feedback and do easier editing in place.  Also looking to add a lot of recipes.  This is moving the documentation farther away from the code, which is admittedly bad, but I think duplication is worse in the short term.  Maybe I can build a generator to build our docs directory off of discourse?  Anyone with admin on discourse can edit the top level docs in response to feedback. 

It is aggressive so what do you guys (@oren @barakmich @dennwc) think?  If cool, go ahead and merge it.


oren:
That's an interesting idea but isn't discourse meant to be a discussion forum? is there a section there that have a wiki-like functionality? What's the the problem we have with using Markdown files or the wiki section on Github?

dennwc:
I'd like to leave Go examples in the repository, because they can be seen
in Godoc, if written correctly.

ср, 16 нояб. 2016 г. в 21:27, oren <notifications@github.com>:

robertmeta:
We could convert those examples to proper Go style examples maybe?  IE: _test packages?

dennwc:
@robertmeta Yeah, that was exactly what I meant.

robertmeta:
@dennwc how about the JS examples / recipes?  The _test format covers the Go side well.

robertmeta:
Will add in a few examples and link them from the docs

dennwc:
@robertmeta JS examples should definitely be on discourse. Most questions are asked there anyway.

One more thing, MD files was also used by Cayley web interface to show docs locally. We should link them to discourse now, or extract them during release build.

robertmeta:
I would like to point them to discourse, let me poke at it.

dennwc:
@robertmeta CI is failing because package examples contains only test files. You can add dummy Go file to make it happy (like [this one](https://github.com/cayleygraph/cayley/blob/master/integration/integration.go)).

robertmeta:
@dennwc they are linked to discourse, once this is squashed merged, I will swap out some of the examples under recipes with pointers to the actual source files.

dennwc:
Need to change links from WebUI, like [this one](https://github.com/cayleygraph/cayley/blob/master/templates/sidebar.tmpl#L51). And we can merge it, I think.

barakmich:
Closing for age, without prejudice. Reopen as necessary.

Issue #469, opened 2016-08-31
Easier Way to Get Full Quads in Results


exincoralius:
If I want to get all information from a query, I have to write something like this:
graph.Vertex("</en/from_the_terrace>", "").Tag("1").Out(["","\u003c/film/film/directed_by\u003e"], "1out2").Tag("2").Out("", "2out3").Tag("3").All();

It would be nice if there were some way to get all the quads related to the query, so that I could make a query like this:
graph.Vertex("</en/from_the_terrace>",).Out("\u003c//film/film/directed_by\u003e").All().Quads();

dennwc:
We are discussing one of the possible solutions for this. Will be implemented a bit later one way or another.

dennwc:
Related [topic](https://discourse.cayley.io/t/paper-chase-reification-aka-putting-data-about-links-on-links/328)

carnalim:
Do we have any insight onto a timeline for a feature like this? We are in need of this as well. We actually talked with Barak in person about this.

dennwc:
It's relatively easy to make it work like this:
```js
p = g.V("node").In("<rdf:subject>") // now we have links!
// save each quad "direction"
p = p.Save("<rdf:subject>","s").Save("<rdf:predicate>","p").Save("<rdf:object>","o")
arr = p.TagArray()
```
But all code path should be carefully examined, because it expects only nodes to be returned, not links. Thus it may crash occasionally if we miss something. This is the only time consuming part that I can think of.
So, If this solution look good for you (and @barakmich), then I can implement it in a few days.

exincoralius:
If you could get it to return both nodes and links, that would be great. Would there be a way of getting labels as well?

dennwc:
Sure. Labels are the same.

carnalim:
Any chance this will be in the 0.6 release?

dennwc:
We already tagged v0.6, so no. I think it will be in a next minor release.

carnalim:
Any Updates on when we might see this?

dennwc:
@carnalim Please test #469. It's missing some reversals and `Has` support, but it works for In/Out/Save.

iddan:
https://github.com/cayleygraph/cayley/issues/820 is a step forwards.

PR #494, opened 2016-11-18
path: Traverse individual quad directions

dennwc:
Work in progress implementation of #469.

Implementation is not checking types yet (nodes vs quads), thus incorrect queries may crash.

PR makes it easier to work with quads in general, for example, save all quads for specific subject:
```js
p = g.V("node").In("<rdf:subject>") // now we have links!
// save each quad "direction"
p = p.Save("<rdf:subject>","s").Save("<rdf:predicate>","p").Save("<rdf:object>","o")
arr = p.TagArray()
```


barakmich:
I'm not sure I'm on board with this. While I can see the intent, I'm not sure we should expose links this way in the language just yet.

There's going to be an overhaul in this vein following reification: with the power to add links on links we'll need a cohesive way to query or even traverse that bit of data, so I'd like to do a bit more design work and  postpone this conversation until after. 

One thing that's useful but not likely to change soon would be a new final step, perhaps .Quads(), which would pre-tag all apparent quad iterators and, for every iteration, call QuadStore.Quad(), to return the subgraph that was traversed in the query.

PR #499, opened 2016-11-23
Updated examples so that they work in repl/http.

mzats:
Replaced angle brackets with quotes in testdata.
Added ".All()" to most examples
Updated some comments about result sets.
Fixed some cut-and-paste errors.

### DIFF
--- a/data/testdata.nq
+++ b/data/testdata.nq
@@ -1,15 +1,15 @@
-"alice" "follows" "bob" .
-"bob" "follows" "fred" .
-"bob" "status" "cool_person" .
-"charlie" "follows" "bob" .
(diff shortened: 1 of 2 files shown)


mzats:
The go tests failed because they rely on the data. Let me know if you'd like me to update the go tests to remove the angle brackets, or if you'd like me to add the angle brackets into the example code.  My vote is for the former, but it's your call.

dennwc:
Angle brackets are correct in the test data, they are part of nquads spec. Tests are also correct.
Thus, Gremlin examples should be changed to include angle brackets where necessary.

mzats:
Fixed, thanks!

PR #512, opened 2016-12-30
Fix build tags

barakmich:
Remove glide Docker dependencies, make Docker an optional dependency, add build tags for docker and appengine tests.


barakmich:
This will speed up our builds drastically; it's a cute idea, and when glide supports tags, we can add it back in the glide file, but docker is a beast of a (transitive) dependency.

PR #514, opened 2017-01-09
Remove extraneous alls

barakmich:
If we're going to remove them anyway, then adding a bunch of alls into a And iterator slows down Optimize() -- so let's just never make them from the query side in the first place.


dennwc:
Maybe it's worth to set duplicate Alls to nil in the slice (except the first), thus next loop in the function will skip these, even if we still have a useful iterator inside.

barakmich:
Done.

PR #515, opened 2017-01-10
Tidy up Optional iterator, and allow MQL to support JS-like string sy…

barakmich:
…ntax


barakmich:
Fix the MQL tests while I'm here to support types and use the generic test infrastructure.

PR #516, opened 2017-01-10
Fix -v option and debug error

barakmich:


graph/proto/helpers.go
```diff
@@ -10,10 +10,10 @@ import (
 // GetNativeValue returns the value stored in Node.
 func (m *NodeData) GetNativeValue() quad.Value {
 	if m == nil {
-		return nil
+		return quad.Raw("")
```
dennwc: Is this change necessary? Might break something.

barakmich: Fair enough. Fixed it the other way.

graph/bolt/iterator.go
```diff
@@ -283,9 +283,18 @@ func (it *Iterator) Size() (int64, bool) {
 }
 
 func (it *Iterator) Describe() graph.Description {
+	name := ""
+	nameOf := it.qs.NameOf(&Token{
+		nodes:  true,
+		bucket: it.bucket,
+		key:    it.checkID,
+	})
+	if nameOf != nil {
```
dennwc: `quad.StringOf` might help :)

graph/proto/helpers.go
```diff
@@ -10,7 +10,7 @@ import (
 // GetNativeValue returns the value stored in Node.
 func (m *NodeData) GetNativeValue() quad.Value {
 	if m == nil {
-		return quad.Raw("")
+		return nil
 	} else if m.Value == nil {
 		if m.Name == "" {
 			return quad.Raw("")
```
dennwc: should also be reverted

PR #529, opened 2017-02-02
Added Tests for Bug Found with Error Deleting a time.Time Quad

henrocdotnet:
Added two simple tests.  One tests reading values back against originals and the other attempts to remove quads.

### DIFF
--- a/graph/graphtest/graphtest.go
+++ b/graph/graphtest/graphtest.go
@@ -49,8 +49,7 @@ func TestAll(t testing.TB, gen DatabaseFunc, conf *Config) {
 		TestDeletedFromIterator(t, gen)
 	}
 	TestLoadTypedQuads(t, gen, conf)
-	TestReadTimeQuadsViaIterator(t, gen, conf)
-	TestRemoveTimeQuadsViaIterator(t, gen, conf)
+	TestRemoveQuadsViaIterator(t, gen, conf)
 	TestAddRemove(t, gen, conf)
 	TestIteratorsAndNextResultOrderA(t, gen)
 	if !conf.UnTyped {
@@ -513,8 +512,15 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
 					pq = quad.Time(time.Unix(seconds, nanos).UTC())
 				}
 			}
+			// Right now it looks like only quad.Time has an Equaler interface.
 			if eq, ok := pq.(quad.Equaler); ok {
 				assert.True(t, eq.Equal(got), "Failed to roundtrip %q (%T), got %q (%T)", pq, pq, got, got)
+				
+				// Built-in time.Equal does not compare timezone values.
+				// Note that asserts.Equal is not being used here as it is causing a panic in the LevelDB backend.
+				if _, ok := pq.(quad.Time); ok {
+					assert.True(t, pq.Native().(time.Time) == got.Native().(time.Time), "Failed to roundtrip %q (%T), got %q (%T)", pq, pq, got, got)
+				}
 			} else {
 				assert.Equal(t, pq, got, "Failed to roundtrip %q (%T)", pq, pq)
 				if !conf.NoHashes {
@@ -528,15 +534,17 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
 	require.Equal(t, int64(7), qs.Size(), "Unexpected quadstore size")
 }
 
-// Came across a bug where a time.Time quad returned by iterator.NewAnd could not be deleted even though it exists.
+// Came across a bug where a time.Time quad could not be deleted even though it exists.
 // dennwc said that this has something to do with protobuf encoding dropping timezone info which is used in hash/node id generation.
-func TestReadTimeQuadsViaIterator(t testing.TB, gen DatabaseFunc, conf *Config) {
+func TestRemoveQuadsViaIterator(t testing.TB, gen DatabaseFunc, conf *Config) {
+	// Locals.
 	qs, opts, closer := gen(t)
 	defer closer()
 
+	// Create writer and test-value set.
 	w := MakeWriter(t, qs, opts)
 	now := time.Now()
-	subject := quad.String("timequad")
+	subject := quad.IRI("timequad")
 	values := []quad.Quad{
 		{subject, quad.String("has_unixtime"), quad.Int(now.UnixNano()), nil},
 		{subject, quad.String("has_time"), quad.Time(now), nil},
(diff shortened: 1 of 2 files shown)


graph/graphtest/graphtest.go
```diff
@@ -526,6 +528,65 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+	now := time.Now()
+	subject := quad.String("timequad")
+	values := []quad.Quad{
+		{subject, quad.String("has_unixtime"), quad.Int(now.UnixNano()), nil},
+		{subject, quad.String("has_time"), quad.Time(now), nil},
+	}
+	err := w.AddQuadSet(values)
+	require.Nil(t, err)
+
+	for _, v := range values {
```
dennwc: Tell me if I'm wrong, but looks like this test will fail even if bug is fixed. It picks one value from test table and makes an iterator which will return two values. So we test value A against both A and B, and then B against the same A and B. It's obvious that one of the checks for each value will fail.
Also, you could remove iteration loop and replace it with `ExpectIteratedQuads` for example.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +528,65 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+	it := iterator.NewAnd(
+		qs,
+		qs.QuadIterator(quad.Subject, qs.ValueOf(subject)),
+	)
+	defer it.Close()
+	for it.Next() {
+		err := w.RemoveQuad(qs.Quad(it.Result()))
+		require.Nil(t, err)
+	}
+
```
dennwc: You should test `it.Err()` here, and maybe check how many quads were removed. Test may skip this loop if iterator is broken and acts like an empty set.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +528,65 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+	now := time.Now()
+	subject := quad.String("timequad")
+	values := []quad.Quad{
+		{subject, quad.String("has_unixtime"), quad.Int(now.UnixNano()), nil},
+		{subject, quad.String("has_time"), quad.Time(now), nil},
+	}
+	err := w.AddQuadSet(values)
+	require.Nil(t, err)
+
+	it := iterator.NewAnd(
```
dennwc: As I mentioned earlier on Slack, this test has nothing to do with `And` iterator. You can safely replace it with `qs.QuadsAllIterator()` and get the same results.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +528,65 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+			assert.Equal(t, v, qs.Quad(it.Result()), "Quads do not match")
+		}
+
+		it.Close()
+	}
+}
+
+// Came across a bug where a time.Time quad returned by iterator.NewAnd could not be deleted even though it exists.
+// dennwc said that this has something to do with protobuf encoding dropping timezone info which is used in hash/node id generation.
+func TestRemoveTimeQuadsViaIterator(t testing.TB, gen DatabaseFunc, conf *Config) {
```
dennwc: Makes sense to generalize it to `TestRemoveQuadsViaIterator` and add some more typed quads from `TestLoadTypedQuads` to test table.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +528,65 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+	qs, opts, closer := gen(t)
+	defer closer()
+
+	w := MakeWriter(t, qs, opts)
+	now := time.Now()
+	subject := quad.String("timequad")
```
dennwc: Please use `quad.IRI` here instead of `quad.String`. RDF only allows IRI and BNodes as subject/predicate, and we may enforce this constraint later.

graph/graphtest/graphtest.go
henrocdotnet: In the check, when I was iterating through the values, I was passing two `QuadIterator`'s to `NewAnd`, one for subject and one for predicate so there should have been only one result returned.  I've moved (am moving) the read / equality check to `TestLoadTypedQuads` though.

graph/graphtest/graphtest.go
```diff
@@ -528,63 +534,38 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
-	it := iterator.NewAnd(
-		qs,
-		qs.QuadIterator(quad.Subject, qs.ValueOf(subject)),
-	)
+	// Create iterator.
+	it := qs.QuadsAllIterator()
 	defer it.Close()
+	
+	// Check iterator size.
+	count, exact := it.Size()
```
dennwc: There is no guarantee that QS supports exact iterator size for All iterator (and any other iterator, in fact), because it can be quite large.
For this case, it's better to just count values in iterator loop.

graph/graphtest/graphtest.go
```diff
@@ -513,8 +512,15 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
 					pq = quad.Time(time.Unix(seconds, nanos).UTC())
 				}
 			}
+			// Right now it looks like only quad.Time has an Equaler interface.
 			if eq, ok := pq.(quad.Equaler); ok {
 				assert.True(t, eq.Equal(got), "Failed to roundtrip %q (%T), got %q (%T)", pq, pq, got, got)
+				
+				// Built-in time.Equal does not compare timezone values.
+				// Note that asserts.Equal is not being used here as it is causing a panic in the LevelDB backend.
+				if _, ok := pq.(quad.Time); ok {
```
dennwc: Well, `quad.Equaler` was created to avoid these direct type assertions, so we should track down the bug in LevelDB instead.
Regarding time zones, we don't store them for now. Time is saved as seconds+nanos in UTC, and are returned in a local timezone (if I recall correctly).

I was trying to say that we can add additional checks at the end of the test that ensures, that it can round-trip values/quads second time.

Bug is the following:
1) Quad is added with some timezone, hash was made from this value on quad insertion.
2) Iterator and QS can still access this time value either from quad/value iterator, or by calling `qs.ValueOf` with original time.
3) Quad returned by QS contains correct hash, but time value is reconstructed in different time zone. Thus, if we try to use this new time value from quad instead of original one, it will fail, since hash is different now.

The fix might be to use UTC for hash calculation. But first you'll need to fix this test :)

graph/graphtest/graphtest.go
```diff
@@ -528,63 +534,38 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
-		it.Close()
-	}
-}
-
-// Came across a bug where a time.Time quad returned by iterator.NewAnd could not be deleted even though it exists.
+// Came across a bug where a time.Time quad could not be deleted even though it exists.
 // dennwc said that this has something to do with protobuf encoding dropping timezone info which is used in hash/node id generation.
-func TestRemoveTimeQuadsViaIterator(t testing.TB, gen DatabaseFunc, conf *Config) {
+func TestRemoveQuadsViaIterator(t testing.TB, gen DatabaseFunc, conf *Config) {
+	// Locals.
```
dennwc: You may omit this these comments, since each test does the same thing.

graph/graphtest/graphtest.go
robertmeta: So, umm, I am a bit confused I guess.  Isn't the use of exact (the 2nd return) umm, to be exact.  If he checks exact first and then count, it should be exact?

dennwc: True. But line below requires result to be exact. This is not true for some backends.

robertmeta: Oh, OK, fair point, not complete coverage.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +527,39 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+	require.Nil(t, err)
+
+	// Create iterator.
+	it := qs.QuadsAllIterator()
+	defer it.Close()
+
+	// Attempt to remove found quads.
+	var count = 0
+	for it.Next() {
+		require.Nil(t, it.Err())
```
dennwc: Place it after the loop.

graph/graphtest/graphtest.go
```diff
@@ -526,6 +527,39 @@ func TestLoadTypedQuads(t testing.TB, gen DatabaseFunc, conf *Config) {
...
+		err := w.RemoveQuad(qs.Quad(it.Result()))
+		require.Nil(t, err)
+		count = count + 1
```
dennwc: Just `count++`.

dennwc:
Fixed in a98837a1258938a75e71e569d8b008f23203735f

PR #531, opened 2017-02-05
Tiny Docker Image

psev:
Reduce the Docker image size from 1.5GB to 38.03.MB.
Uses alpine:3.5 as a base container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,34 @@
-FROM golang:latest
-MAINTAINER Barak Michener <me@barakmich.com>
+FROM alpine:3.5
 
-# Set up workdir
-WORKDIR /go/src/github.com/cayleygraph/cayley
+ENV INSTALL_PATH="/usr/local/bin"
 
-# Restore vendored dependencies
-RUN sh -c "curl https://glide.sh/get | sh"
-ADD glide.* ./
-RUN glide install
+ENV PKG="ca-certificates" \
+    PKG_TMP="g++ git go glide" \
+    PKG_CACHE="/var/cache/apk"
 
-# Add and install cayley
-ADD . .
-RUN go install -v ./cmd/cayley
+ENV BUILD_LIB="github.com/cayleygraph/cayley" \
+    BUILD_TAG="master"
 
-# Expose the port and volume for configuration and data persistence. If you're
-# using a backend like bolt, make sure the file is saved to this directory.
-RUN mkdir /data #VOLUME ["/data"]
-EXPOSE 64210
+ENV GOPATH="/go"
 
-RUN echo '{"database":"bolt","db_path":"/data/cayley","read_only":false}' > /data/cayley.cfg
-RUN cayley init -db bolt -dbpath /data/cayley -quads ./data/30kmoviedata.nq.gz
+# running all these steps together builds a smaller image
+RUN mkdir -p ${GOPATH} && \
+    apk update && \
+    apk add ${PKG} && \
+    apk add ${PKG_TMP} && \
+    go get ${BUILD_LIB} && \
+    cd ${GOPATH}/src/${BUILD_LIB} && \
+    git checkout ${BUILD_TAG} && \
+    glide install && \
+    go install -v ./cmd/cayley && \
+    cp -a ${GOPATH}/bin/* ${INSTALL_PATH} && \
+    rm -rf ${GOPATH} && \
+    rm -rf /root/.glide && \
+    apk del --purge ${PKG_TMP} && \
+    rm -rf ${PKG_CACHE}/*
 
-ENTRYPOINT ["cayley", "http", "-config", "/data/cayley.cfg", "-host", "0.0.0.0", "-port", "64210"]
+ENV PATH=${INSTALL_PATH}
+
+EXPOSE [ "64210" ]
+
+ENTRYPOINT [ "/usr/local/bin/cayley" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,4 +36,4 @@ RUN mkdir -p ${GOPATH} && \
     apk del --purge ${PKG_TMP} && \
     rm -rf ${PKG_CACHE}/*
 
-ENTRYPOINT [ "/usr/local/bin/cayley", "-assets", "/assets" ]
+ENTRYPOINT [ "/usr/local/bin/cayley" ]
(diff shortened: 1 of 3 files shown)


barakmich:
This is rather nice; and it looks like it just builds on Quay. Right now we mount in `/data` for datafiles and config; this can clearly still do that, but overall I'll request a matching change for the docs on how to use this Docker image. As for the image itself, it looks good :)

psev:
Should I list myself as the maintainer?

barakmich:
If you intend to maintain it, sure :) My only ask is that you file a bug if you decide not to do so anymore.

psev:
Yeah man, I don't mind at all. Just wanted to verify.

psev:
I've updated the container documentation as per your request @barakmich.

Also, I split up the container into a general use container (~40MB) and the quickstart container (~600MB). The quickstart container has the 30kmovidata preloaded.

To get started with Cayley, all you need to do is run:

```
docker run -it --rm -p 64210:64210 sugarush/cayley-quickstart
```

Dockerfile
```diff
@@ -1,24 +1,39 @@
...
+    cp -a templates /assets && \
+    glide install && \
+    go install -v ./cmd/cayley && \
+    cp -a ${GOPATH}/bin/* ${INSTALL_PATH} && \
+    rm -rf ${GOPATH} && \
+    rm -rf /root/.glide && \
+    apk del --purge ${PKG_TMP} && \
+    rm -rf ${PKG_CACHE}/*
+
+ENTRYPOINT [ "/usr/local/bin/cayley" ]
```
dennwc: Add assets and host to entrypoint?

Dockerfile
```diff
@@ -1,24 +1,39 @@
...
+ENV PKG="ca-certificates" \
+    PKG_TMP="g++ git go glide" \
+    PKG_CACHE="/var/cache/apk"
 
-# Expose the port and volume for configuration and data persistence. If you're
-# using a backend like bolt, make sure the file is saved to this directory.
-RUN mkdir /data #VOLUME ["/data"]
-EXPOSE 64210
+ENV BUILD_LIB="github.com/cayleygraph/cayley" \
+    BUILD_TAG="master"
```
dennwc: This should depend on current branch for CI to work properly.

Dockerfile
```diff
@@ -1,24 +1,39 @@
...
-RUN sh -c "curl https://glide.sh/get | sh"
-ADD glide.* ./
-RUN glide install
+ENV INSTALL_PATH="/usr/local/bin"
 
-# Add and install cayley
-ADD . .
-RUN go install -v ./cmd/cayley
+ENV PKG="ca-certificates" \
+    PKG_TMP="g++ git go glide" \
```
dennwc: What version of Go is it? Can we have a parameter for it?

Dockerfile-quickstart
```diff
@@ -0,0 +1,18 @@
+FROM sugarush/cayley
```
dennwc: It's a good idea to mirror an image to `quay.io/cayleygraph/cayley`, thus this `FROM` statement might be different.

psev: I planned on changing it to Quay, though personally I avoid Quay and deploy my own registries.

Dockerfile
psev: I think for the generic container, it should be left up to the user. Those options are set for the quickstart container.

dennwc: I think setting assets does not hurt for generic use case.

Setting host to anything else in Docker is nearly useless. The only reason might be networking in host mode, which is should not be set for DB.

In both cases user is free to set custom entrypoint if he has a good reason for it.

Dockerfile
psev: This uses Go 1.7.3. I don't think a parameter for Go version will be feasible.

The version of Go is determined by the version of Alpine Linux.

Alpine 3.5 == Go 1.7.3
Alpine 3.4 == Go 1.6.3
Alpine 3.3 == Go 1.5.4

Dockerfile
psev: Docker builds should be decoupled from the source code. This allows that exactly.

dennwc: Why exactly? It's useful to make CI build for master and dev branches + all of the release tags. Each of those Docker images can be then pushed with version/branch tags. Hardcoding branch  will break this approach. Alternatively, `sed` may be used to switch branch and commit in Dockerfile according to one being tested in CI environment.

Dockerfile
dennwc: I think it's important to have a direct control over it, because it affects performance (1.7 was one good example) and stdlib features we can use (recent includes: context, plugins, ...). Go version should be pined the same way as we pin other dependencies.

Dockerfile
psev: The issue I ran into is/was the assets flag can't be set before the command is specified, i.e. http, repl, etc.

You're completely right about the host being anything other than the wildcard address for listen address inside a container, but the previous statement still applies to the host flag.

I could easily fix this with github.com/spf13/cobra, not sure if it's a good fit though.

Dockerfile
psev: All that is needed to change to one of the Go versions listed above is to include *FROM alpine:3.x* in the Dockerfile.

Dockerfile
dennwc: Yeah, forgot about other commands in this context, sorry :)

We already have a cobra-based implementation, but we are planning to cut a minor release before merging it since it changes some flag names.

Dockerfile
psev: Could you point me to a list of the Travis CI build environment variables?

Dockerfile
dennwc: I'd better add `wget+tar -xz` of exact version mention in ENV, for example. I don't think that there is a guarantee that they preserve an exact version for each release (and how often they will).

Dockerfile
psev: Awesome.

Dockerfile
dennwc: Sure: https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

dennwc:
New CLI is merged, so we can revive this one. @psev are you still interested in maintaining this image, or should I rewrite it?

dennwc:
Closing in favor of https://github.com/cayleygraph/cayley/pull/637

PR #534, opened 2017-02-12
CockroachDB transaction retries

mikaelcabot:
After using the `cockroach` flavor for some time, I have encountered the 40001 `retry transaction` error a lot. CockroachDB recommends Client-Side transaction retries.

https://www.cockroachlabs.com/docs/transactions.html#client-side-intervention
https://www.cockroachlabs.com/docs/transactions.html#client-side-transaction-retries

### DIFF
--- a/graph/sql/cockroach.go
+++ b/graph/sql/cockroach.go
@@ -82,7 +82,48 @@ type AmbiguousCommitError struct {
 	error
 }
 
+// runTxCockroach runs the transaction and will retry in case of a retryable error.
+// https://www.cockroachlabs.com/docs/transactions.html#client-side-transaction-retries
 func runTxCockroach(tx *sql.Tx, in []graph.Delta, opts graph.IgnoreOpts) error {
+	// Specify that we intend to retry this txn in case of CockroachDB retryable
+	// errors.
+	if _, err := tx.Exec("SAVEPOINT cockroach_restart"); err != nil {
+		return err
+	}
+
+	for {
+		released := false
+
+		err := tryRunTxCockroach(tx, in, opts)
+
+		if err == nil {
+			// RELEASE acts like COMMIT in CockroachDB. We use it since it gives us an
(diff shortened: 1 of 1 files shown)


dennwc:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[graph/sql/cockroach.go, line 110 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/534#-KcnUIFtQt68krB7vGkz:-KcnUIFtQt68krB7vGl-:b-a7nuie) ([raw file](https://github.com/cayleygraph/cayley/blob/87ba1a0d36e1e835322ce7aab96ac0712fdf853d/graph/sql/cockroach.go#L110)):*
> ```Go
> 	)
> 
> 	runTx := func() error {
> ```

Semantics of this function are not clear to me. It is meant to be retry-able, but it's not re-initializing variables above.

Does this means that Cockroach still stores transaction data? When why are we writing the same data again? And if it's not, why `inserted` is still initialized with data from previous TX? It will not insert nodes data in this case.

It also makes sense to make it a separate function `runTxCockroach` and rename an old one to `retryTxCockroach`, or similar.

---

*[graph/sql/cockroach.go, line 228 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/534#-KcnU48JFnEbodopTZzd:-KcnU48Km9AHtzx-D9Nr:b11i90y) ([raw file](https://github.com/cayleygraph/cayley/blob/87ba1a0d36e1e835322ce7aab96ac0712fdf853d/graph/sql/cockroach.go#L228)):*
> ```Go
> 	)
> 	for {
> 		released = false
> ```

`released := false`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/534)*

dennwc:
Thanks for working on this! But there are potentials problems in this PR.

dennwc:
Looks good, thanks!

PR #536, opened 2017-02-13
Clean up

Yannic:


Yannic:
Thanks for reviewing.
I just checked why this condition was triggered and found out that pkgPath is always 'github.com/cayleygraph/cayley/graph'. I'm not sure if it is possible to use reflect without changing QuadStoreRegistration

dennwc:
Ok, I see the problem now.

We can solve it by adding `Type interface{}` field to `QuadStoreRegistration` to store nil QS pointer, so later we can find a package name, for example.

Another solution would be to make `NewFunc` an `interface{}` and check signature at runtime, but this is not a best approach.

Yannic:
I don’t think having a `Type interface{}` or changing `NewFunc` to `interface{}` would be a good solution.
I’d propose to change `QuadStoreRegistration` to
```go
type QuadStoreRegistration interface {
	// Create new QuadStore
	// Must return ErrOperationNotSupported if operation is not supported
	New(string, Options) (QuadStore, error)

	// Create new QuadStore for request
	// Must return ErrOperationNotSupported if operation is not supported
	NewForRequest(QuadStore, Options) (QuadStore, error)

	// Upgrade existing QuadStore
	// Must return ErrOperationNotSupported if operation is not supported
	Upgrade(string, Options) error

	// Initialize new QuadStore
	// Must return ErrOperationNotSupported if operation is not supported
	Init(string, Options) error

	// QuadStore is persistent
	// Must return ErrOperationNotSupported if operation is not supported
	IsPersistent() bool
}
```
I know this would mean to require all QuadStores to implement this interface, but they also have change if we add `Type interface{}` to `QuadStoreRegistration`.

dennwc:
Having `Type interface{}` is a backward-compatible change. We can omit adding PkgPath in this case (and panic if QuadStoreRegistration has NewForRequest set, since it's the only method that uses PkgPath).

For your proposed change we have to break user's code, thus it cannot be merged to v0.6.1 and have to wait until v0.7. Also, we are changing one small boilerplate (`Type()` method) to a much larger boilerplate (definition of struct that implements QuadStoreRegistration). I don't think it's worth it.

The fun thing about zero pointer is that we can keep Type method on QS, and even call it on this zero ptr.

Yannic:
I undid the change of `QuadStoreRegistration`. 
All other changes do not break anything and can be merged now.

graph/quadstore.go
```diff
@@ -125,119 +125,52 @@ type QuadStore interface {
 
 type Options map[string]interface{}
 
+var (
+	typeInt = reflect.TypeOf(0)
```
dennwc: `int(0)`, just in case

graph/quadstore.go
```diff
@@ -125,119 +125,52 @@ type QuadStore interface {
...
+
 func (d Options) IntKey(key string) (int, bool, error) {
 	if val, ok := d[key]; ok {
-		switch vv := val.(type) {
-		case float64:
-			return int(vv), true, nil
-		default:
-			return 0, false, fmt.Errorf("Invalid %s parameter type from config: %T", key, val)
+		if reflect.TypeOf(val).ConvertibleTo(typeInt) {
+			return val.(int), true, nil
```
dennwc: This will panic. You are type-asserting `float64` to `int`, not converting it.

graph/quadstore.go
```diff
@@ -125,119 +125,52 @@ type QuadStore interface {
...
+
+		return false, false, fmt.Errorf("Invalid %s parameter type from config: %T", key, val)
 	}
+
 	return false, false, nil
 }
 
-var ErrCannotBulkLoad = errors.New("quadstore: cannot bulk load")
-var ErrDatabaseExists = errors.New("quadstore: cannot init; database already exists")
+var ErrCannotBulkLoad = fmt.Errorf("quadstore: cannot bulk load")
```
dennwc: Leave it as `errors.New`. It's not a good practice to use `fmt.Errorf` if you don't need to.

graph/registry.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func RegisterQuadStore(name string, register QuadStoreRegistration) {
+	if register.NewFunc == nil {
+		panic("NewFunc must not be nil")
+	}
+
+	// Register QuadStore with friendly name
+	if _, found := storeRegistry[name]; found {
+		panic(fmt.Sprintf("Already registered QuadStore \"%s\".", name))
```
dennwc: `%q` instead of `\"%s\"`

graph/registry.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func UpgradeQuadStore(name string, dbpath string, opts Options) error {
+	r, registered := storeRegistry[name]
+	if !registered {
+		return ErrQuadStoreNotRegistred
+	}
+
+	if r.UpgradeFunc == nil {
+		return ErrOperationNotSupported
```
dennwc: Not sure about this one. It's should just return nil, I think.

Issue #286, opened 2015-07-26
Using LevelDB with cayley: Error, could not open! file does not exist


ZhengYang:
Configuration file:

```
$ cat cayley.cfg
{
"database": "leveldb",
"db_path": "/cayley_0.4.1_darwin_386/db/",
"read_only": false
}
```

Command: 

```
./cayley repl -ignoredup -dbpath=/wikidata-simple-statements.nt -config=/cayley_0.4.1_darwin_386/cayley.cfg
```

Error:

```
E0726 22:56:50.286284 81645 quadstore.go:118] Error, could not open!  file does not exist
E0726 22:56:50.288093 81645 cayley.go:251] file does not exist
```

ds--:
The config file takes priority over command line, I don't think the -dbpath switch will do anything in your case since the config file has db_path specified.

From the error, I assumed "/cayley_0.4.1_darwin_386/db/" has not been initialized. Was ``./cayley init -config=/cayley_0.4.1_darwin_386/cayley.cfg`` executed prior to running cayley repl?

jeffw-wherethebitsroam:
I had the same issue (for both bolt and leveldb), and running init resolved it. But it would be great if the error message was a bit more helpful! The leveldb message above doesn't give any clue about the issue, bolt just panics with the message:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

carnalim:
I am having this exact same issue with Level, I don't have a config file created either just using command line parameters: 
./cayley http --db="leveldb" --dbpath=/tmp/leveldb

dennwc:
Since e849da94022cc444921a11b1baabef13ebd7cb10 you need to init database first:
`./cayley init --db="leveldb" --dbpath=/tmp/leveldb`
or
`./cayley http --db="leveldb" --dbpath=/tmp/leveldb --init`

This behavior should be documented better. Ping @robertmeta.

robertmeta:
I think this also needs to be fixed if command line doesn't override config, as I think that is unexpected behavior.

healiseu:
Hi, I am a newcomer and I met this error. It is better I think if you make clear at the beginning of your Quickstart guide that in order to play with any backend you must run these three commands in sequence (initialization, loading, backend), e.g.

1. `./cayley init --db=bolt --dbpath=/tmp/test`
2. `./cayley load --db=bolt --dbpath=/tmp/test --quads=testdata.nq --alsologtostderr`
[...]

PR #538, opened 2017-02-15
Issue #286 init db and config documentation

jtorvald:
Hopefully made the quickstart application a bit more clear for running init and then load/repl.
Also did a small fix of returning a more specific error message when a database file can not be opened. Not sure if this is necessary since there is a new CLI on its way.
And explained the binding on localhost and host 0.0.0.0 if people try to access it on a virtual machine / test-server.

### DIFF
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -49,7 +49,7 @@ func Open(cfg *config.Config) (*graph.Handle, error) {
 
 func OpenQuadStore(cfg *config.Config) (graph.QuadStore, error) {
 	clog.Infof("Opening quad store %q at %s", cfg.DatabaseType, cfg.DatabasePath)
-	qs, err := graph.NewQuadStore(cfg.DatabaseType, cfg.DatabasePath, cfg.DatabaseOptions);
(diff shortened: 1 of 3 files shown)


internal/db/db.go
```diff
@@ -49,6 +50,12 @@ func Open(cfg *config.Config) (*graph.Handle, error) {
 func OpenQuadStore(cfg *config.Config) (graph.QuadStore, error) {
 	clog.Infof("Opening quad store %q at %s", cfg.DatabaseType, cfg.DatabasePath)
 	qs, err := graph.NewQuadStore(cfg.DatabaseType, cfg.DatabasePath, cfg.DatabaseOptions)
+
+	// override error to make it more informative
+	if os.IsNotExist(err) {
+		err = errors.New(fmt.Sprintf("file does not exist: %s. Please use with --init or run ./cayley init when it is a new database (see docs for more information)", cfg.DatabasePath))
```
dennwc: Please use `fmt.Errorf` here.

internal/db/db.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	"github.com/cayleygraph/cayley/graph"
 	"github.com/cayleygraph/cayley/internal/config"
+	"os"
```
dennwc: Move to the top.

PR #551, opened 2017-04-05
add nds caching for gaedatastore Get, Put operations

alainbrown:


dennwc:
@alainbrown Thanks for trying this caching library with Cayley. Do you have any benchmarks to see how it affects performance?

Also, please remove docker from glide.lock. We intentionally removed it to save time for CI builds.

alainbrown:
Thanks @dennwc I'll go ahead and made the changes you mentioned to the glide.lock file.

As far as benchmarks, do you have any suggestions for capturing that? Or maybe an example of something folks might find interesting here?

For my purpose, it is mainly for cost saving as google datastore charges on a per operation basis but the memcached layer operations are free. Each query on my graph touch hundreds of nodes, but since a lot of them are the same nodes and quads again and again, we get a huge cost savings.

dennwc:
@alainbrown Results of integration tests will be useful. Just want to make sure nothing gone wrong with a new library.

wsh:
I wonder if this is a logical next step for my work on #656. `nds` (the library this adds) depends on running in App Engine Standard, which makes sense because that's an existing constraint for that backend. Presumably caching might be useful at the `nosql` meta-backend layer. In any case, _that_ caching work should depend on memcached/redis/some OSS protocol rather than the proprietary App Engine Memcache service.

dennwc:
@wsh Sounds reasonable. Closing this in favor of new backend implementation and caching at nosql layer.

PR #554, opened 2017-04-11
mongo: reorder tx operations

dennwc:
Reorder tx operations to avoid broken quads for concurrent readers.


tbayne:
Looks good hear, this should resolve the issue of broken quads being returned because nodes are not written before the quad itself.

Good job on adding a test for this too!

PR #560, opened 2017-04-16
gizmo: generate docs from Go function comments; resolve #520

### LINKED ISSUE #520: gizmo: generate md documentation from function comments
Since all Gizmo functions are registered automatically via reflection, we can use comments in Go code to generate documentation for these functions.
This will improve documentation in Go code, and will keep Gizmo docs in sync with an actual code (or at least comments to the code).

dennwc:
First iteration of automatic docs generator for Gizmo.


barakmich:
Just one thought: a `GremlinAPI.md.in`, such that we can wrap the generated docs. And maybe some badge suggesting the doc is generated, and therefore way more likely to be up-to-date than not.

PR #567, opened 2017-04-19
cockroach: schema changes within transactions are restricted

mikaelcabot:
Added a function CreateSQLTables on flavor struct for optional "flavor specific" creation of tables/indexes.

Cockroach beta-20170413 release notes:
https://www.cockroachlabs.com/docs/beta-20170413.html#sql-language-changes

### DIFF
--- a/graph/sql/cockroach.go
+++ b/graph/sql/cockroach.go
@@ -15,97 +15,65 @@ import (
 const flavorCockroach = "cockroach"
 
 func init() {
-	nodesTable := `CREATE TABLE nodes (
-		hash BYTEA PRIMARY KEY,
-		value BYTEA,
-		value_string TEXT,
-		datatype TEXT,
-		language TEXT,
-		iri BOOLEAN,
-		bnode BOOLEAN,
-		value_int BIGINT,
-		value_bool BOOLEAN,
-		value_float double precision,
-		value_time timestamp with time zone,
-		FAMILY fhash (hash),
-		FAMILY fvalue (value, value_string, datatype, language, iri, bnode,
-			value_int, value_bool, value_float, value_time)
-	);`
-
-	quadsTable := `CREATE TABLE quads (
-		horizon BIGSERIAL PRIMARY KEY,
-		subject_hash BYTEA NOT NULL,
-		predicate_hash BYTEA NOT NULL,
-		object_hash BYTEA NOT NULL,
-		label_hash BYTEA,
-		id BIGINT,
-		ts timestamp
-	);`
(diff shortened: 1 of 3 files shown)


graph/sql/quadstore.go
```diff
@@ -111,6 +111,8 @@ type Flavor struct {
 	Error       func(error) error
 	Estimated   func(table string) string
 	RunTx       func(tx *sql.Tx, in []graph.Delta, opts graph.IgnoreOpts) error
+	// CreateSQLTables for a flavor specific creation of tables
+	CreateSQLTables *func(conn *sql.DB, options graph.Options) error
```
dennwc: function is already a pointer, you should not make it a pointer again

graph/sql/cockroach_test.go
```diff
@@ -3,34 +3,75 @@
...
+				},
+			},
 		}
-		conn.Close()
-		return true
-	})
-	db, err := connect(`postgresql://root@`+addr+`:26257?sslmode=disable`, flavorPostgres, nil)
+
+		host = "localhost"
+		_, closer = dock.RunAndWait(t, conf, func(string) bool {
```
dennwc: This is the same code block as on line 62. It needs to be a single code path with different variable values.

Also, you are doing this trick for darwin for SQL only. How about Mongo or any other non-trivial backend? This code really needs to be in `dock.RunAndWait` instead.

graph/sql/cockroach.go
```diff
@@ -15,64 +15,97 @@ import (
 const flavorCockroach = "cockroach"
 
 func init() {
+	nodesTable := `CREATE TABLE nodes (
```
dennwc: `const`?

graph/sql/cockroach.go
```diff
@@ -15,64 +15,97 @@ import (
...
+		switch e.Code {
+		case "42P07":
+			return graph.ErrDatabaseExists
+		}
+		return err
+	}
+
+	// Schema changes within transactions are restricted in CockroachDB
+	// See: https://www.cockroachlabs.com/docs/beta-20170413.html#sql-language-changes
+	createSQLTables := func(conn *sql.DB, options graph.Options) (err error) {
```
dennwc: I thinks it's not specific enough to be a thing unique for Cockroach. Maybe it's better to add a bool flag that turns this behavior for any SQL backend? Like `NoSchemaChangesInTx`. It will allow us to reuse this flag later if we decide to add indexes at runtime, for example. What do you think?

PR #579, opened 2017-04-29
Glog issues

MarioAriasGa:
Some glog changes have broken repl and deprecated documentation.

### DIFF
--- a/cmd/cayley/command/repl.go
+++ b/cmd/cayley/command/repl.go
@@ -64,7 +64,7 @@ func NewReplCmd() *cobra.Command {
 	}
 	langs := query.Languages()
 	cmd.Flags().Bool("init", false, "initialize the database before using it")
-	cmd.Flags().StringP("lang", "l", "gizmo", `query language to use ("`+strings.Join(langs, `", "`)+`")`)
+	cmd.Flags().StringP("lang", "", "gizmo", `query language to use ("`+strings.Join(langs, `", "`)+`")`)
 	cmd.Flags().DurationP("timeout", "t", 30*time.Second, "elapsed time until an individual query times out")
 	registerLoadFlags(cmd)
 	return cmd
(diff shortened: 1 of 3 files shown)


cmd/cayley/command/repl.go
```diff
@@ -64,7 +64,7 @@ func NewReplCmd() *cobra.Command {
 	}
 	langs := query.Languages()
 	cmd.Flags().Bool("init", false, "initialize the database before using it")
-	cmd.Flags().StringP("lang", "l", "gizmo", `query language to use ("`+strings.Join(langs, `", "`)+`")`)
+	cmd.Flags().StringP("lang", "", "gizmo", `query language to use ("`+strings.Join(langs, `", "`)+`")`)
```
dennwc: `cmd.Flags().String("lang", "gizmo", ...`

docs/Quickstart-As-Application.md
```diff
@@ -49,7 +49,7 @@ After the database is initialized we load the data.
 And wait. It will load. If you'd like to watch it load, you can run
 
 ```bash
-./cayley load -c cayley_overview.yml -i data/testdata.nq --alsologtostderr
+./cayley load -c cayley_overview.yml -i data/testdata.nq --logtostderr=true
```
dennwc: Please also change remap at `cmd/cayley/cayley.go:172`

cmd/cayley/command/repl.go
MarioAriasGa: Changed, thanks.

docs/Quickstart-As-Application.md
MarioAriasGa: It seems that alsologtostderr is still in glog, but the remap wasn't working. I think that the entry for alsologtostderr in [cayley.go:170](https://github.com/cayleygraph/cayley/blob/master/cmd/cayley/cayley.go#L170) was trying to avoid passing the boolean for convenience, but that wasn't working for me.

PR #584, opened 2017-05-02
Introduce the new Bolt backend

barakmich:
Rewritten from the ground up, now with fewer, smaller indexes. Notable changes:
* New `bolt2` backend
* Updated all protos

Should completely match functionality with the current bolt backend, while being much easier to extend. It passes our conformance tests, deletion and recreation.

Things that need followup:
* Compaction is not yet finished, so deletion still takes space in the log. This could be a broader interface question, however.
* It should a backend option for the size of the bloom filter to use. The current setting is optimized for O(100M) quads. Fortunately, this option is only a performance enhancement for load speed; if this map becomes saturated, loads will be slower, but functionally equivalent. 

Some comparisons:

Initialization:
```
$ time ./cayley init -d bolt -a b1 
I0502 16:08:56.702345   11835 cayley.go:68] Cayley version: 0.7.0-alpha (dev snapshot)
I0502 16:08:56.702519   11835 database.go:180] using backend "bolt" (b1)
./cayley init -d bolt -a b1  0.01s user 0.00s system 8% cpu 0.082 total
$ time ./cayley init -d bolt2 -a b2
I0502 16:09:01.949057   11857 cayley.go:68] Cayley version: 0.7.0-alpha (dev snapshot)
I0502 16:09:01.949296   11857 database.go:180] using backend "bolt2" (b2)
./cayley init -d bolt2 -a b2  0.53s user 0.12s system 10% cpu 6.266 total
$ du b1 b2
24K	b1
6820K	b2
```
The slower init time and increased size is due to doing some small but constant (at about 6 MB/6 seconds) work up front. The win here comes with the loading:
```
$ time ./cayley load -d bolt -a b1 -i data/30kmoviedata.nq.gz 
I0502 16:12:04.959106   11969 cayley.go:68] Cayley version: 0.7.0-alpha (dev snapshot)
I0502 16:12:04.959284   11969 database.go:180] using backend "bolt" (b1)
./cayley load -d bolt -a b1 -i data/30kmoviedata.nq.gz  24.70s user 2.79s system 25% cpu 1:48.79 total
$ time ./cayley load -d bolt2 -a b2 -i data/30kmoviedata.nq.gz
I0502 16:14:06.343962   12071 cayley.go:68] Cayley version: 0.7.0-alpha (dev snapshot)
I0502 16:14:06.344257   12071 database.go:180] using backend "bolt2" (b2)
I0502 16:14:06.345922   12071 indexing.go:482] Using a bloom filter of 77951328 bytes
./cayley load -d bolt2 -a b2 -i data/30kmoviedata.nq.gz  8.72s user 0.71s system 30% cpu 30.695 total
$ du b1 b2
418840K	b1
105596K	b2
```

Integration test speeds:

```
$ benchx bolt1.txt bolt2.txt
benchmark                                     old ns/op     new ns/op     delta	mult
BenchmarkNamePredicate-4                      145283        116473        -19.83%	-1.25x
BenchmarkLargeSetsNoIntersection-4            8935835       454970        -94.91%	-19.65x
BenchmarkVeryLargeSetsSmallIntersection-4     22526524      811584        -96.40%	-27.78x
BenchmarkHelplessContainsChecker-4            37178830      202289        -99.46%	-185.19x
BenchmarkHelplessNotContainsFilms-4           37725220      410579        -98.91%	-91.74x
BenchmarkHelplessNotContainsActors-4          39732456      422515        -98.94%	-94.34x
BenchmarkNetAndSpeed-4                        1334512       687601        -48.48%	-1.94x
BenchmarkKeanuAndNet-4                        772429        586583        -24.06%	-1.32x
BenchmarkKeanuAndSpeed-4                      1013624       575575        -43.22%	-1.76x
BenchmarkKeanuOther-4                         5711807       790977        -86.15%	-7.22x
BenchmarkKeanuBullockOther-4                  9701172       1067793       -88.99%	-9.08x
BenchmarkSaveBogartPerformances-4             355276        173124        -51.27%	-2.05x
```

I maintain that the 100x wins are due to query non-optimization cases that we haven't yet determined; that needs to be a followup as well. The 2x cases, however, are likely believable.

This means the old bolt, as well as some other backends, have the potential to see some big wins if we track down why. But we can still enjoy some speed while we figure it out :)


graph/bolt2/indexing.go
```diff
@@ -0,0 +1,519 @@
...
+		link.Timestamp = time.Now().UnixNano()
+		err = qs.index(tx, &link, nil)
+		writePrimToBuf(&link)
+		qs.exists.Add(bloomBuf)
+		if err != nil {
+			return err
+		}
+		qs.size++
+	}
+	err = qs.flushMapBucket(tx)
```
dennwc: This map has the same scope as ApplyDelta, if I'm not mistaken. Make it a local variable here?

barakmich: If I did, I'd have to pass them everywhere, which I wanted to avoid.

graph/bolt2/quadstore.go
```diff
@@ -0,0 +1,326 @@
...
+		return err
+	})
+	if err != nil {
+		clog.Errorf("error fetching quad %d: %s", key, err)
+	}
+	return v
+}
+
+func (qs *QuadStore) primitiveToQuad(tx *bolt.Tx, p *proto.Primitive) (quad.Quad, error) {
+	q := &quad.Quad{}
```
barakmich: This is the quad we're filling in. We have to allocate it. (can't quad.Set nil)

barakmich:
Fixed all, except the ones with good reason.

graph/bolt2/quadstore.go
dennwc: Right, but it can be just `var q quad.Quad`, and you can avoid `*q` in each return. Not a big deal though.

graph/bolt2/all_iterator.go
```diff
@@ -0,0 +1,204 @@
...
+func (it *AllIterator) Contains(v graph.Value) bool {
+	if it.nodes {
+		x, ok := v.(Int64Value)
+		if !ok {
+			return false
+		}
+		it.id = uint64(x)
+		return it.id <= uint64(it.horizon)
+	}
+	it.prim = v.(*proto.Primitive)
```
dennwc: `, ok  = `

barakmich:
NextSequence could work, but it's an odd place for it. Easier to just handle it ourselves (plus it's one less thing in the generic KV world). 

Fixed the check.

PR #585, opened 2017-05-04
Shape: New query optimizer

dennwc:
This change is a rewrite of https://github.com/cayleygraph/cayley/pull/368 and defines a new representation of query shapes that can be a target of generic query optimizations. Each query shape has an equivalent iterator (or a tree of iterators) and should be built to run a query, similar to how path lib works.

**Some bonuses we get:**

1) Easier to work with query shapes: no more `Kind()`, `SubIterators()[1]`, `Close()`, `Clone()`, etc. Just Go types with public fields that are discoverable with reflect.
2) Query shapes are not bound to backend. Generic optimizations can be done before building an actual iterator tree and results of optimization can be reused.
3) Since these structs does not store any state, they can be serialized and restored later. Useful for pagination of complex queries, stored procedures, etc.
4) Backends like SQL can place their own query shapes during optimization process, thus it would be easier to work with - it's still a small Go structs with few public fields.
5) These structs are usually cheaper to create than real iterators.
6) They are extremely easy to debug: simply use `Print` to get the whole tree out.

**What's is done:**

1) Most iterators have a corresponding Shape. Any Shape can be used to build an iterator tree for any QuadStore.
2) Path lib converted to use shapes instead of iterators under the hood with no API change.
3) Some simple optimizations are ported from iterators to shapes. Benchmarks show -7% to +4% difference on integration tests. I'm considering this a random noise, since no new optimizations are done in this layer yet.

**What's next:**

Optimize high-level shapes in QuadStores as a follow-up. This is especially interesting to do for SQL. KV backends such as Bolt will benefit as well, because we can do more in the same View transaction knowing the exact query upfront. Consider this simple example:
```golang
HasA{Dir:Subject, Quads:{
    QuadFilter{Dir: Object, Values: Lookup{"Humphrey Bogart"} },
    QuadFilter{Dir: Predicate, Values: Lookup{"name"} },
}}
```
This shape contains 6 iterators. But the query is actually much simpler (after optimization pass):
```golang
QuadsAct{
    Result: Subject, Save: map[Direction][]string{},
    Filter:map[Direction]graph.Value{
        Object: 221036, Predicate: 185834,
    },
}
```
First thing to notice is that names are already gone because optimization does a values lookup upfront (in batches, potentially).

Second, `QuadsAct` construct is one of the high-level iterators I mentioned. It represents an iterator that visits each quad which falls under Filter constraints (using object-predicate-subject index in this case), tags directions using Save, and returns a value of Result direction as a result of iterator. This can be done in one pass in every backend we support.

And we can do even more with this :)

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/585)


barakmich:
Alright, I'm still digesting, but here are some surface-level comments on the first pass. One thing that would help a ton, both for reviewing and usability would be some more documentation comments in the new package. At the very least, all the `structs` and their members

The thing that sells me on the underlying notion is
```
Intersect{
	AllNodes{},
	QuadDirection{Dir: quad.Subject, Quads: Quads{}},
	Intersect{
		Lookup{quad.IRI("alice")},
		Unique{QuadDirection{Dir: quad.Object, Quads: Quads{}}},
	},
},
```

 I mean, that's just nice :smile: 

If I'm also following, the idea is for all queries to start by creating the shapes first, optimizing, and then filling with iterators just before evaluation?

That works fairly well. Are there exceptions to that?

Another general thing I'd like to see -- just a proof-of-concept in the test -- is to serialize one to a string and read it back again, then evaluate it.

---

Reviewed 6 of 11 files at r1.
Review status: 6 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[graph/bolt/bolt_test.go, line 211 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD10_oago4LKqs3_Rn:-KkD10_oago4LKqs3_Ro:b-zd3gzo) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/bolt/bolt_test.go#L211)):*
> ```Go
> }
> 
> func TestBoltShape(t *testing.T) {
> ```

Add to some other backends, esp. memory and maybe bolt2?

---

*[graph/iterator/materialize.go, line 187 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD0Y5lzdWngzuNo3TH:-KkD0Y5lzdWngzuNo3TI:b-uc7cz8) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/iterator/materialize.go#L187)):*
> ```Go
> 	)
> 	if it.expectSize > 0 {
> 		size, exact = it.expectSize, true
> ```

s/true/false/

---

*[graph/shape/path.go, line 22 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD4m2ObkZJ9BMAcCHP:-KkD4m2ObkZJ9BMAcCHQ:b-wqtcru) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/shape/path.go#L22)):*
> ```Go
> }
> 
> func Start(nodes ...graph.Value) Path {
> ```

Switch the name for Start and StartValues? Most external packages that want shapes will want Start() to take quad.Values and, if the internal package really wants graph.Values, that should call StartValues

---

*[graph/shape/shape_test.go, line 169 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD4O-gt3Bu1mzeXHKc:-KkD4O-gt3Bu1mzeXHKd:b-sku0bj) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/shape/shape_test.go#L169)):*
> ```Go
> }
> 
> type lookupQuadStore map[quad.Value]graph.Value
> ```

Separate file for this and the interface impl below? Maybe there's a place for some of these mock quadstores to live together? (There's another one in the codebase somewhere)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-:-KkD5XWb2FFjkj0pr6gi:bkc1dmn)*

dennwc:
> If I'm also following, the idea is for all queries to start by creating the shapes first, optimizing, and then filling with iterators just before evaluation?
>
> That works fairly well. Are there exceptions to that?

Exactly! The idea is that shapes can be optimized "offline", sometimes even without quadstore. Because of this shapes can be reused and cached by application, serialized, etc. But they are even more usable with quadstore, since SQL can define it's own set of operations, for example. No need to hack through iterators kinds and states.

The only exception from this rule is a path lib which defines an Iterator shape for compatibility (it clones iterator on Build). All other shapes are "immutable", stateless and are not bound to any backend.

Although, there are few special cases - some shapes can be execute during optimization step (if quadstore is available during optimization). For example, Lookup shape runs a ValueOf for fixed set of values, and since in most cases this array is small, optimizer tries to lookup all values *before* iteration. This helps to resolve predicates earlier, and use some additional information such as the size of indexes to optimize query even further. This usually helps to jump from name nodes to object nodes, and use new stats to evaluate a final query. It can resolve up to 3-5 levels of shapes on integration tests this way.

> Another general thing I'd like to see -- just a proof-of-concept in the test -- is to serialize one to a string and read it back again, then evaluate it.

It's just a matter of using `encoding/gob` package :) Most of the time I print shapes via `%#v` and paste the result as a var in a test case. So I can state that it works.

But to make it usable by others, I think we need to define a protobuf or JSON serialization. This might take some time, so let's do it as a follow-up.

---

Review status: 6 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[graph/iterator/materialize.go, line 187 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD0Y5lzdWngzuNo3TH:-KkDFtWutCU_MWv3RTn2:b-ksaicy) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/iterator/materialize.go#L187)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

s/true/false/
</blockquote></details>

It's a bit subtle.

Field is called `expectedSize` because this size was reported as exact by one of quad indexes at the time of shape optimization. So it can be used with a high level of confidence for Stats, and it should affect iterator optimization as well. This is why `exact` is set to true.

On the other hand, some quadstores does not provide a snapshot of data, so we can never know if index is still of that size or not.

I guess we need three states: RoughlyEstimated (just a guess), NearlyExact (size from a past observation) and Exact (exactly what an iterator will return).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-:-KkDEFCHvCG522S-9xDV:b-q5bqo3)*

barakmich:
Review status: 6 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[graph/iterator/materialize.go, line 187 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD0Y5lzdWngzuNo3TH:-KkDIwThL90GX6vcdRUS:bjiwjkp) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/iterator/materialize.go#L187)):*

It's a bit subtle.

Field is called `expectedSize` because this size was reported as exact by one of quad indexes at the time of shape optimization. So it can be used with a high level of confidence for Stats, and it should affect iterator optimization as well. This is why `exact` is set to true.

On the other hand, some quadstores does not provide a snapshot of data, so we can never know if index is still of that size or not.

I guess we need three states: RoughlyEstimated (just a guess), NearlyExact (size from a past observation) and Exact (exactly what an iterator will return).
</blockquote></details>

I'm good with that enum, though I think that's a separate change. It's probably `false` today, because it may not be exact.

The docs on that enum are interesting
RoughlyEstimated // just a guess, at worst should return an upper bound (eg, the size of the database)
NearlyExact // an educated guess, from past observation or quick computation (eg, compressed size multiplied by average compression)
Exact // yep, you'll get N.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

dennwc:
Added more comments.

---

Review status: 6 of 11 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[graph/bolt/bolt_test.go, line 211 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD10_oago4LKqs3_Rn:-KkDOMPL5Ub3rGmRC26U:b8k50d0) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/bolt/bolt_test.go#L211)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Add to some other backends, esp. memory and maybe bolt2?
</blockquote></details>

Bolt2 added and memstore is already in `shape_test.go`.

---

*[graph/iterator/materialize.go, line 187 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD0Y5lzdWngzuNo3TH:-KkDOKVu9YHw3EYNwc3F:bv9h0m7) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/iterator/materialize.go#L187)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

I'm good with that enum, though I think that's a separate change. It's probably `false` today, because it may not be exact.

The docs on that enum are interesting
RoughlyEstimated // just a guess, at worst should return an upper bound (eg, the size of the database)
NearlyExact // an educated guess, from past observation or quick computation (eg, compressed size multiplied by average compression)
Exact // yep, you'll get N.

</blockquote></details>

Reverted to `false` for now.

---

*[graph/shape/path.go, line 22 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD4m2ObkZJ9BMAcCHP:-KkDP81MYmG9HzyIoNmZ:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/shape/path.go#L22)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Switch the name for Start and StartValues? Most external packages that want shapes will want Start() to take quad.Values and, if the internal package really wants graph.Values, that should call StartValues
</blockquote></details>

Done.

---

*[graph/shape/shape_test.go, line 169 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkD4O-gt3Bu1mzeXHKc:-KkDWh1myMnb8yAwvgFm:b-nxsexk) ([raw file](https://github.com/cayleygraph/cayley/blob/e57134cbb3bfb0ac5cf0d6388e6ed5c99301353d/graph/shape/shape_test.go#L169)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Separate file for this and the interface impl below? Maybe there's a place for some of these mock quadstores to live together? (There's another one in the codebase somewhere)
</blockquote></details>

Done, see `graphmock` package.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-:-KkDj4SNUv4knUDTfjeF:b-b575vw)*

barakmich:
Some more things; mostly naming.

---

Reviewed 2 of 11 files at r1, 20 of 22 files at r2.
Review status: 26 of 28 files reviewed at latest revision, 5 unresolved discussions.

---

*[graph/shape/shape.go, line 121 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvlQqBz6WTynSXTPqd:-KkvlQqBz6WTynSXTPqe:bhuk6m8) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L121)):*
> ```Go
> type Except struct {
> 	Nodes Shape // nodes to exclude
> 	All   Shape // a set of all nodes to exclude from; nil means AllNodes
> ```

`Exclude` and `From` instead of `Nodes` and `All`? `From` is similar to below and Nodes is a bit general.

---

*[graph/shape/shape.go, line 246 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kkvm5bbucgZe7p2tkCi:-Kkvm5bcipvO4wa5OiL1:b-x2j332) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L246)):*
> ```Go
> // QuadFilter is a constraint used to filter quads that have a certain set of values on a given direction.
> // Analog of LinksTo iterator.
> type QuadFilter struct {
> ```

It's not filtering them, though, it's fanning them out; it's the first half of a traversal. So maybe a better name here; `QuadsFor`? Or `EdgesFor`? And then `QuadDirection` below could be `NodesFor`.  Because the common shape is "Get the QuadsFor these nodes, and then the NodesFor these quads" (or `NodesFrom`?)

---

*[graph/shape/shape.go, line 420 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvnahrzFLz6qdg7HyK:-KkvnahrzFLz6qdg7HyL:b6euo0q) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L420)):*
> ```Go
> // and returns a specified quad direction as result of the iterator (equivalent of HasA).
> // Optionally, Size field may be set to indicate an approximate number of quads that will be returned by this query.
> type QuadsAct struct {
> ```

Maybe `QuadAction`? Or `QuadsAction`? No harm in making it more explicit.

---

*[graph/shape/shape.go, line 512 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kkvnude2LVaqblqJKfl:-Kkvnude2LVaqblqJKfm:b-u9iu48) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L512)):*
> ```Go
> 
> // Fixed is a static set of nodes. Defined only for a particular QuadStore.
> type Fixed []graph.Value
> ```

While `append` would work, maybe it makes sense to have an `Add` method for this type, so we don't have to remember that it's a slice underneath? Similarly to the other types like this (`Quads` above, `Lookup` below) -- while we can instantiate them as, eg, `Lookup{val1, val2}` it can't hurt to have it be available programmatically ('for x in list, add to lookup')

---

*[graph/shape/shape.go, line 601 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvoevSnDribFSaZo2r:-KkvoevTxDOeEGSLoT0M:b-dhqdc6) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L601)):*
> ```Go
> }
> 
> var MaterializeThreshold = 100 // TODO: tune
> ```

Maybe the struct should have a threshold field? Not sure how we'd initialize it to something by default

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-:-Kkvpmw-qkC3sKt97s4C:b-9tese4)*

dennwc:
Review status: 26 of 28 files reviewed at latest revision, 6 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kl5UU3YcR-wEHTFcI-R:-Kl5UU3YcR-wEHTFcI-S:b-62bvm5) (no related file):*
@barakmich What do you think about `shape/path.go`? I've made it before an integration with path lib, and now I see that it nearly useless. We can use good old path lib to make these shapes without changing anything. Or maybe there is some value in having a path-like specifically for shapes?

---

*[graph/shape/shape.go, line 121 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvlQqBz6WTynSXTPqd:-Kl5OkTZ7qQPHEXXnD0B:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L121)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

`Exclude` and `From` instead of `Nodes` and `All`? `From` is similar to below and Nodes is a bit general.
</blockquote></details>

Done.

---

*[graph/shape/shape.go, line 246 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kkvm5bbucgZe7p2tkCi:-Kl5RBg5kmB_OjZ5af7Q:byw7tma) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L246)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

It's not filtering them, though, it's fanning them out; it's the first half of a traversal. So maybe a better name here; `QuadsFor`? Or `EdgesFor`? And then `QuadDirection` below could be `NodesFor`.  Because the common shape is "Get the QuadsFor these nodes, and then the NodesFor these quads" (or `NodesFrom`?)
</blockquote></details>

Yeah, `NodesFrom` is much better. `QuadDirection` was a terrible one.

But I'm not sure about `QuadsFor`. I made buildIterator on this type private to encourage usage of Quads with multiple constraints/filters, so optimizer can do less work joining them together. Because of this, `QuadFilter` only appears in context of Quads. It's more self-descriptive to see a slice of `QuadFilters` there (meaning is clear: intersect these filters to get quads), instead of slice of `QuadsFor` structs (it intersects them, or makes a union?).

---

*[graph/shape/shape.go, line 420 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvnahrzFLz6qdg7HyK:-Kl5P0YrFxnwX-0OrqNj:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L420)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Maybe `QuadAction`? Or `QuadsAction`? No harm in making it more explicit.
</blockquote></details>

Done.

---

*[graph/shape/shape.go, line 512 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kkvnude2LVaqblqJKfl:-Kl5Q6McoaeU-L5fNBmZ:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L512)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

While `append` would work, maybe it makes sense to have an `Add` method for this type, so we don't have to remember that it's a slice underneath? Similarly to the other types like this (`Quads` above, `Lookup` below) -- while we can instantiate them as, eg, `Lookup{val1, val2}` it can't hurt to have it be available programmatically ('for x in list, add to lookup')
</blockquote></details>

Done.

---

*[graph/shape/shape.go, line 601 at r2](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KkvoevSnDribFSaZo2r:-KkwAKLVjem4x1J6qnHq:b-xnqc99) ([raw file](https://github.com/cayleygraph/cayley/blob/8954224a50e24375596071aafe0befef1eaf4dc5/graph/shape/shape.go#L601)):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Maybe the struct should have a threshold field? Not sure how we'd initialize it to something by default
</blockquote></details>

It's more like an global option for optimizer, it doesn't act as a limit for individual Materialize iterators/shapes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

dennwc:
@barakmich Rebased, PTAL

barakmich:
Reviewed 1 of 22 files at r2, 9 of 9 files at r3.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kl5UU3YcR-wEHTFcI-R:-KqzSzwmwfX6nwl9SYwx:brozfjw) (no related file):*

@barakmich What do you think about `shape/path.go`? I've made it before an integration with path lib, and now I see that it nearly useless. We can use good old path lib to make these shapes without changing anything. Or maybe there is some value in having a path-like specifically for shapes?
</blockquote></details>

Yeah, I like that idea. Use the old path lib, and then call the new Shape() on it

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KqzTtBwbvZb_RuY7Wa0:-KqzTtBxLAXZGITdzS5p:b-ql4apu) (no related file):*
Double checking; the current backend optimizers still run as always, and it's a followup to make them operate on shapes (and eventually deprecate a bunch of repeated code with them)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

dennwc:

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kl5UU3YcR-wEHTFcI-R:-Kr-ve7Gc5qO4g8YzvY0:b-a7yghl) (no related file):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Yeah, I like that idea. Use the old path lib, and then call the new Shape() on it
</blockquote></details>

Should I drop `shape/path.go` then? It nearly mirrors the code in `path`.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-KqzTtBwbvZb_RuY7Wa0:-Kr-uYycTjil052RwS9T:baaao8z) (no related file):*
<details><summary><i>Previously, barakmich (Barak Michener) wrote…</i></summary><blockquote>

Double checking; the current backend optimizers still run as always, and it's a followup to make them operate on shapes (and eventually deprecate a bunch of repeated code with them)?
</blockquote></details>

Yes, I haven't touched backend optimizers yet. But the shape optimizer works already. So it currently optimizes things multiple times: for shapes, for iterators and finally for the backend.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

barakmich:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/585#-Kl5UU3YcR-wEHTFcI-R:-Kr30RCLTwT9nHpy-LK8:bpnhkjy) (no related file):*

Should I drop `shape/path.go` then? It nearly mirrors the code in `path`.
</blockquote></details>

Yeah, go for it

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

dennwc:
@barakmich Removed path lib mirror and squashed.

barakmich:
Reviewed 4 of 7 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/585)*

barakmich:
LGTM

PR #586, opened 2017-05-05
Variables WIP

BlakeMScurr:
This is a work in progress that would allow queries with 'variables' such as:
```
v := NewVar()
p := StartPath().Out().Is(v).Out().Has("<follows>", v)
```

This particular query will match all cases where a node is followed by a node with the same value. Generally, variables allow distinct parts of the query to rely on one another. More exposition of the idea on [discourse](https://discourse.cayley.io/t/graph-shaped-queries/686).

Alter the iterator interface to pass context through Next(), NextPath(), and Contains(). This gives us a shared object through which to sync variable values.

Create a variable iterator that either acts as the source of the values for the variable, or just grabs uses the current one from the context object (bind/use). 

Reorder the iterator tree so that binders only have their next method called, and users only have their Contains() method called.

Stop using the materialize iterator, and clear the contains map of the recursive iterator on variable value change - contains() can return different results at different times depending on variables, and should not be cached.


dennwc:
@BlakeMScurr Will you continue to work on this?

PR #587, opened 2017-05-06
Update transaction.go

h4ck3rm1k3:
spelling

### DIFF
--- a/graph/transaction.go
+++ b/graph/transaction.go
@@ -16,7 +16,7 @@ package graph
 
 import "github.com/cayleygraph/cayley/quad"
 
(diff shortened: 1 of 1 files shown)


barakmich:
First one should be 'atomically'

h4ck3rm1k3:
OK, I was wondering about that, but found almost no usage of that word online. How about using a longer description?

barakmich:
Sure! Go for it!

h4ck3rm1k3:
see second commit.

barakmich:
Eh, better enough.

PR #595, opened 2017-05-14
Update sidebar.tmpl

h4ck3rm1k3:
overview is not there

### DIFF
--- a/templates/sidebar.tmpl
+++ b/templates/sidebar.tmpl
@@ -49,7 +49,7 @@
         <button class="btn btn-sm center-block dropdown-toggle" type="button" data-toggle="dropdown"> Documentation <span class="caret"></span>
         </button>
         <ul class="dropdown-menu">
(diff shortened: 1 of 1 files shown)


templates/sidebar.tmpl
```diff
@@ -49,7 +49,7 @@
         <button class="btn btn-sm center-block dropdown-toggle" type="button" data-toggle="dropdown"> Documentation <span class="caret"></span>
         </button>
         <ul class="dropdown-menu">
-            <li><a href="/docs/Overview" target="_blank">Overview</a></li>
+            <li><a href="/docs/Overview" target="_blank">Quickstart-As-Application</a></li>
```
dennwc: You wanted to change `href` instead of text?

h4ck3rm1k3: good point

dennwc:
Thanks :)

PR #603, opened 2017-05-15
Fix example startup command

chrispassas:
Update example start up command. The existing example results in an error. It appears --port was added as a separate argument later on.


dennwc:
Thanks! I forgot to update in at the time of change.

PR #634, opened 2017-09-21
ElasticSearch as Supported Backend

michaelqiu94:
- Iterator and QuadStore implementations for ElasticSearch
- Basic optimizations

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/634)


dennwc:
First review pass. More to come :)

---

Reviewed 1 of 2 files at r1, 3 of 4 files at r5, 1 of 1 files at r6, 1 of 1 files at r7.
Review status: 6 of 8 files reviewed at latest revision, 21 unresolved discussions.

---

*[cayley.db, line 0 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuaa3SLGqsSBZB9xAit:-Kuaa3SMSSLZHaGEznk4:b-rvwju7) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/cayley.db#L0)):*
Please remove this binary blob from PR.

---

*[cmd/cayley/cayley.go, line 39 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuaaL3KOeqWLh8BSFPg:-KuaaL3KOeqWLh8BSFPh:b-edkcij) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/cmd/cayley/cayley.go#L39)):*
> ```Go
> 	_ "github.com/cayleygraph/cayley/graph/bolt"
> 	_ "github.com/cayleygraph/cayley/graph/bolt2"
> 	_ "github.com/cayleygraph/cayley/graph/elastic"
> ```

Add this import to integration tests as well.

---

*[graph/elastic/iterator.go, line 21 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuaaQkS95pR86veMMMi:-KuaaQkS95pR86veMMMj:bi04m0u) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/iterator.go#L21)):*
> ```Go
> 	"github.com/cayleygraph/cayley/graph/iterator"
> 	"github.com/cayleygraph/cayley/quad"
> 	"golang.org/x/net/context"
> ```

Use `"context"` import from stdlib.

---

*[graph/elastic/iterator.go, line 50 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuaeyHx7uB0W2DpaNFW:-KuaeyHx7uB0W2DpaNFX:b-k4baon) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/iterator.go#L50)):*
> ```Go
> }
> 
> var ctx = context.Background()
> ```

Do not use global context variables at package level.
I know this is a repetition, but you should pass this explicitly to all calls that require it. We will pass real contexts soon enough.

---

*[graph/elastic/iterator.go, line 176 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuaj7MOLVIlFCtNEvWx:-Kuaj7MOLVIlFCtNEvWy:bp3hrun) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/iterator.go#L176)):*
> ```Go
> // Reset makes a result set
> func (it *Iterator) Reset() {
> 	results, err := it.makeElasticResultSet()
> ```

Set the `it.resultSet` to `nil` and `Next` will do `it.makeElasticResultSet` for you automatically

---

*[graph/elastic/quadstore.go, line 26 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuaaoWHPArxWm6Xdktl:-KuaaoWHPArxWm6Xdktm:b51pnc3) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L26)):*
> ```Go
> 	"time"
> 
> 	elasticClient "gopkg.in/olivere/elastic.v5"
> ```

Other files refer to it as `elastic`.

---

*[graph/elastic/quadstore.go, line 97 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuaay1ZfX_cLpQSEfiM:-Kuaay1ZfX_cLpQSEfiN:b-f3q4ha) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L97)):*
> ```Go
> 	client, err := elasticClient.NewClient(elasticClient.SetURL(addr))
> 	if err != nil {
> 		panic(err)
> ```

return this error instead of panicking

---

*[graph/elastic/quadstore.go, line 141 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuajV48W3GBN9Rn1-Hl:-KuajV48W3GBN9Rn1-Hm:b-eu5ylo) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L141)):*
> ```Go
> 				},
> 				"active": {
> 					"type": "string"
> ```

`boolean`? same for nodes

---

*[graph/elastic/quadstore.go, line 184 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuahGPqMOT-vAzO76cS:-KuahGPqMOT-vAzO76cT:b-fgg7j2) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L184)):*
> ```Go
> 	}
> 	qs.client = client
> 	qs.client.OpenIndex("cayley")
> ```

Shouldn't the result be saved in struct, or something? I don't think it's doing anything useful right now.

---

*[graph/elastic/quadstore.go, line 250 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuabStmNLlr2a9ClEfg:-KuabStmNLlr2a9ClEfh:bn9unpt) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L250)):*
> ```Go
> 	src, err := query.Source()
> 	if err != nil {
> 		panic(err)
> ```

return this error

---

*[graph/elastic/quadstore.go, line 258 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuabUugYJ2rnTOVt0zN:-KuabUugYJ2rnTOVt0zO:bb7od0r) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L258)):*
> ```Go
> 	key, err := json.Marshal(src)
> 	if err != nil {
> 		panic(err)
> ```

same

---

*[graph/elastic/quadstore.go, line 267 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuacDLNrCHIOfr12wtA:-KuacDLNrCHIOfr12wtB:btwwqq0) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L267)):*
> ```Go
> func (qs *QuadStore) checkValid(key string) bool {
> 	// Check if a quad with that key already exists (meaning it is a duplicate delta). If so, return true.
> 	get1, err := qs.client.Get().Index("cayley").Type("quads").Id(key).Do(context.Background())
> ```

`get`, `got` or `res`

---

*[graph/elastic/quadstore.go, line 271 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuac8hvRwz8ilrk8Rwa:-Kuac8hvRwz8ilrk8Rwb:b-4aq0uc) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L271)):*
> ```Go
> 		return false
> 	}
> 	if get1.Found {
> ```

`return get1.Found`

---

*[graph/elastic/quadstore.go, line 279 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuaiCyqhxWCbG38UOka:-Kuafsukd99EvfslVCH-:bdhs08r) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L279)):*
> ```Go
> 
> // ElasticNodeTracker - Keeping track of nodes in a quad for graph deletes
> type ElasticNodeTracker struct {
> ```

This should not be exported by the package

---

*[graph/elastic/quadstore.go, line 280 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuafpAbwJh6oHOi6ibH:-KuaciqoEEBPXjk5zS9k:b-bwzzxq) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L280)):*
> ```Go
> // ElasticNodeTracker - Keeping track of nodes in a quad for graph deletes
> type ElasticNodeTracker struct {
> 	NodeType  string
> ```

`quad.Direction`?

---

*[graph/elastic/quadstore.go, line 322 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuahdr2GBm8rZulJyxN:-Kuahdr2GBm8rZulJyxO:b-ezjzp9) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L322)):*
> ```Go
> 		// Keeps track of the graph action (add or delete) for each kind of node (sub, pred, obj, label).
> 		// If d.Action is add, add the node. Else, delete it.
> 		nodeTracker[d.Quad.Subject] = ElasticNodeTracker{
> ```

What if this node is in other direction ("type" in this code) of some quad in this transaction?

---

*[graph/elastic/quadstore.go, line 409 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuadX-YITNunaU34F80:-KuadX-YITNunaU34F81:b-l4o9ro) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L409)):*
> ```Go
> 
> 		if err != nil {
> 			clog.Errorf("Error: %v", err)
> ```

Print in caller instead (or do not print and pass the value to it's caller, if possible)

---

*[graph/elastic/quadstore.go, line 445 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuae6FjORn4TaAU9oje:-Kuae6FjORn4TaAU9ojf:b12y62e) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L445)):*
> ```Go
> 			Query(termQuery).
> 			From(0).Size(1).
> 			Pretty(true).
> ```

Is `pretty=true` necessary?

---

*[graph/elastic/quadstore.go, line 455 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuadxIYj7vJvpX3s7ua:-KuadxIYj7vJvpX3s7ub:b4yddek) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L455)):*
> ```Go
> 		if searchResult.Hits.TotalHits > 1 {
> 			return nil
> 		} else {
> ```

Omit `else` and lower the code below

---

*[graph/elastic/quadstore.go, line 575 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuageWGVpm0JTK3yI4C:-KuageWGVpm0JTK3yI4D:b-bf2as1) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L575)):*
> ```Go
> 	}
> 
> 	if searchResult.Hits.TotalHits > 0 {
> ```

```golang
if searchResult.Hits.TotalHits == 0 {
  return nil
}
// rest of the code follows
```

---

*[graph/elastic/quadstore.go, line 578 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuak7ZFGPDL672mikln:-Kuak7ZFGPDL672miklo:bcn04hd) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L578)):*
> ```Go
> 		hit := searchResult.Hits.Hits[0]
> 		// convert json to object
> 		var eNode ElasticNode
> ```

I don't think they are compatible with `ElasticNodeEntry`. You should use the same process for storing and retrieving values. Mongo, for example, converts nodes on insert.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-Kualk7B4HiZRJtcnlmZ:b-3xhtid)*

dennwc:
Reviewed 2 of 4 files at r8.
Review status: 6 of 8 files reviewed at latest revision, 11 unresolved discussions.

---

*[graph/elastic/quadstore.go, line 184 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuahGPqMOT-vAzO76cS:-KuiuanG0E6vYgdrnuEa:bsuerpm) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L184)):*

Shouldn't the result be saved in struct, or something? I don't think it's doing anything useful right now.
</blockquote></details>

I mean, OpenIndex does literally nothing here. It return a proxy struct but it's discarded. None requests are sent to the server as a part of this call, as far as I know.

---

*[graph/elastic/quadstore.go, line 271 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuac8hvRwz8ilrk8Rwa:-Kuiv9Uo7OGTAxQrIu7s:bjengv7) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L271)):*

`return get1.Found`
</blockquote></details>

Yo may omit parentheses

---

*[graph/elastic/quadstore.go, line 445 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuae6FjORn4TaAU9oje:-KuivQIqMjDVD6z-3POd:b4tbx6w) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L445)):*

Is `pretty=true` necessary?
</blockquote></details>

Same for other places

---

*[graph/elastic/quadstore.go, line 447 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KuivV54Wj5rpm8vAznf:-KuivV54Wj5rpm8vAzng:b-dteitw) ([raw file](https://github.com/cayleygraph/cayley/blob/7264762ae5ed1279a934ac8cf523818fedfca877/graph/elastic/quadstore.go#L447)):*
> ```Go
> 
> 		// Delete Node from nodes Type in elasticsearch
> 		_, deleteErr := qs.client.Delete().
> ```

It's okay to call it `err` and use `=`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
@dennwc or @barakmich, could you review the changes one more time and let us know if it looks good? Thanks!

dennwc:
Some comments are still unresolved. And sure, I'll take another look in following days.

---

Review status: 6 of 8 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-KwA51wS-qYN2f707Xms:boxdzuk)*

camazotz:
@dennwc Addressed the remaining comments and just pushed. Replying to two of your comments below.

First comment:

> graph/elastic/quadstore.go, line 322 at r7 (raw file):
> 
> // Keeps track of the graph action (add or delete) for each kind of node (sub, pred, obj, label).
> 		// If d.Action is add, add the node. Else, delete it.
> 		nodeTracker[d.Quad.Subject] = ElasticNodeTracker{
> 
> What if this node is in other direction ("type" in this code) of some quad in this transaction?

The deltas come in as Quads, so there are bound to be all four directions (subject, predicate, object, label) with each one. If one of the nodes already exists as a direction in another quad, Elasticsearch won't insert it again since it's a duplicate.

Second comment:

> graph/elastic/quadstore.go, line 578 at r7 (raw file):
> 
> hit := searchResult.Hits.Hits[0]
> 		// convert json to object
> 		var eNode ElasticNode
> 
> I don't think they are compatible with ElasticNodeEntry. You should use the same process for storing and retrieving values. Mongo, for example, converts nodes on insert.

We are using ElasticNode for storing and retrieving values (same process). Just like Mongo, we convert nodes on insert (elastic/quadstore.go line 452):

```
doc := ElasticNodeEntry{
			Hash: nodeId,
			Node: nodeVal,
		}

		// Add document to index
		_, err := qs.client.Index().Index(DefaultESIndex).Type("nodes").Id(nodeId).BodyJson(doc).Do(context.Background())
```

So ElasticNodeEntry is only used while inserting in the updateNodeBy method.

dennwc:
If it's possible, please reply via Reviewable, it will keep comment threads separate for each discussion. Thanks.

---

Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 322 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuahdr2GBm8rZulJyxN:-KwiqiXZ5zrI4u2Qrqi8:b-xr28bc) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L322)):*
> The deltas come in as Quads, so there are bound to be all four directions (subject, predicate, object, label) with each one. If one of the nodes already exists as a direction in another quad, Elasticsearch won't insert it again since it's a duplicate.

I should have explained the case a bit better. What I meant is that currently `nodeTracker` asserts that there is only one action per node, and it can be only in one direction. Consider this case:

```
ADD <a> <b> <c>
DEL <b> <d> <a> 
```

In current code it will record an add operation first (pseudocode):
```
tracker[<a>] = {Subject, Add}
tracker[<b>] = {Predicate, Add}
...
tracker[<b>] = {Subject, Del}
...
tracker[<a>] = {Object, Del}
```
Two problems to note:
- Since value is the key, it will first be written as `<a> = Subject`, and then `<a> = Object`. So in final map there will be no records of `<a>` being in `Subject`. And if you don't differentiate between these cases, why store a direction in the first place?
- Again, since value is a key in the map, the first operation marks `<a> = Add`, while second one marks it as `Del`. If I understood correctly, the node will be removed in the process, but it shouldn't since it's used in one of quads in this Delta (and maybe in other quads in DB). The correct behavior will be to track how many times node was added (+1) and deleted (-1). The final value should be added to existing record in the database, and if it becomes 0 in the process, this will mean that value should be removed.

---

*[graph/elastic/quadstore.go, line 578 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuak7ZFGPDL672mikln:-Kwj2LbX6rbspYwLy0VD:bswetde) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L578)):*
> So ElasticNodeEntry is only used while inserting in the updateNodeBy method.

This is exactly the problem. There is nothing wrong with struct itself (Mongo [does the same](https://github.com/cayleygraph/cayley/blob/master/graph/mongo/quadstore.go#L519)), but there is a difference in how Mongo and Elastic store values. Currently you rely on Mongo's code to read values, but you don't convert values on insert, as [Mongo does](https://github.com/cayleygraph/cayley/blob/master/graph/mongo/quadstore.go#L205). This will most certainly lead to Time values encode incorrectly, and maybe some others.

---

*[graph/elastic/quadstore.go, line 130 at r10](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kwj10KzFCp2ryCwNf_K:-Kwj10KzFCp2ryCwNf_L:b-12ndbs) ([raw file](https://github.com/cayleygraph/cayley/blob/12e7e40509dd48484fc53625dfabdc374712c6dd/graph/elastic/quadstore.go#L130)):*
> ```Go
> 			"properties": {
> 				"subject": {
> 					"type": "string"
> ```

It should be `keyword`, since hashes can be searched only by exact value. Same for others.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-Kwj3x2PPYJdknkrGEns:b-7xwbae)*

camazotz:
Review status: 6 of 8 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 322 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuahdr2GBm8rZulJyxN:-Kwl3KM2ZEFS0u0Y27mG:b-l8z461) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L322)):*

> The deltas come in as Quads, so there are bound to be all four directions (subject, predicate, object, label) with each one. If one of the nodes already exists as a direction in another quad, Elasticsearch won't insert it again since it's a duplicate.

I should have explained the case a bit better. What I meant is that currently `nodeTracker` asserts that there is only one action per node, and it can be only in one direction. Consider this case:

```
ADD <a> <b> <c>
DEL <b> <d> <a> 
```

In current code it will record an add operation first (pseudocode):
```
tracker[<a>] = {Subject, Add}
tracker[<b>] = {Predicate, Add}
...
tracker[<b>] = {Subject, Del}
...
tracker[<a>] = {Object, Del}
```
Two problems to note:
- Since value is the key, it will first be written as `<a> = Subject`, and then `<a> = Object`. So in final map there will be no records of `<a>` being in `Subject`. And if you don't differentiate between these cases, why store a direction in the first place?
- Again, since value is a key in the map, the first operation marks `<a> = Add`, while second one marks it as `Del`. If I understood correctly, the node will be removed in the process, but it shouldn't since it's used in one of quads in this Delta (and maybe in other quads in DB). The correct behavior will be to track how many times node was added (+1) and deleted (-1). The final value should be added to existing record in the database, and if it becomes 0 in the process, this will mean that value should be removed.
</blockquote></details>

The direction is stored because it is needed in the updateNodeBy method. If the delta is a graph delete, I check if the Node already exists in another quad. In order to construct the Elasticsearch query, both the direction and the value are needed (your first point).

[quadstore.go, line 459](https://github.com/michaelqiu94/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L459)

```
case graph.Delete:

		// Construct an Elastic query to check if the Node marked for deletion is present
		// in another Quad. If so, don't delete the Node. If not, delete the Node.
		// Example: nodeType - subject, nodeId - 9328afb
		termQuery := elastic.NewTermQuery(trackedNode.NodeType.String(), nodeId)
		ctx := context.Background()

		// Elasticsearch query checking the quads Type
		searchResult, err := qs.client.Search().
			Index(DefaultESIndex).
			Type("quads").
			Query(termQuery).
			From(0).Size(1).
			Do(ctx)
		if err != nil {
			return err
		}

		// If the Node is present in more than one Quad, don't delete
		if searchResult.Hits.TotalHits > 1 {
			return nil
		}
```

Performing this lookup prevents the Node from being deleted if it is used in another quad (your second point). I saw the counter tracking approach in mongo but if cayley is run in parallel (multiple threads or multiple servers), there could be synchronization issues with the count. So we figured a lookup would be the safest bet.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
I should take another look at how iterators work, and run our tests against Elastic.

---

Review status: 6 of 8 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 322 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuahdr2GBm8rZulJyxN:-KwoKMqVt0BJhB2CZmRH:b-hodbu6) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L322)):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

The direction is stored because it is needed in the updateNodeBy method. If the delta is a graph delete, I check if the Node already exists in another quad. In order to construct the Elasticsearch query, both the direction and the value are needed (your first point).

[quadstore.go, line 459](https://github.com/michaelqiu94/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L459)

```
case graph.Delete:

		// Construct an Elastic query to check if the Node marked for deletion is present
		// in another Quad. If so, don't delete the Node. If not, delete the Node.
		// Example: nodeType - subject, nodeId - 9328afb
		termQuery := elastic.NewTermQuery(trackedNode.NodeType.String(), nodeId)
		ctx := context.Background()

		// Elasticsearch query checking the quads Type
		searchResult, err := qs.client.Search().
			Index(DefaultESIndex).
			Type("quads").
			Query(termQuery).
			From(0).Size(1).
			Do(ctx)
		if err != nil {
			return err
		}

		// If the Node is present in more than one Quad, don't delete
		if searchResult.Hits.TotalHits > 1 {
			return nil
		}
```

Performing this lookup prevents the Node from being deleted if it is used in another quad (your second point). I saw the counter tracking approach in mongo but if cayley is run in parallel (multiple threads or multiple servers), there could be synchronization issues with the count. So we figured a lookup would be the safest bet.

</blockquote></details>

Thanks for clarification, I forgot about this code path. An approach is totally valid.
There might be performance concerns, but Elastic has pretty good indexing, so it might not be a problem.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-KwoLRrYOXM4pSyeRr8d:b47tbf7) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*
> ```Go
> 		return string(d)
> 	case quad.TypedString:
> 		return string(d.Value)
> ```

This will loose the type entirely. Same for language strings.
We usually store a secondary field that saves the type or language.
But if you want, we can assume for now that Elastic does not support types and make it a separate PR. Although it will fail most tests with current approach.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-KwoMGwFfv2Tgz4jzTqt:b-zcs6ly)*

dennwc:
Sorry for the delay. Most test are OK, assuming current comments are addressed.
But few fail because of incorrect value encoding (as expected).

Please check `graph/graphtest/graphtest.go` and specifically these tests:
`TestLoadTypedQuads`
`TestCompareTypedValues`

Reason they fail is that we have few types of values like IRI, BNode, LangString and TypedString. Test tries to write them to QuadStore and expects htat it can load values back without any modifications. But it fails because currently you value conversion function wipes any type information and converts all values to plain strings.  This can be addressed by storing values in the following struct:

```golang
type Node struct{
   ID string
   Value string // text
   IRI bool
   BNode bool
   Lang string
   Type string
}
```

Thus, `toElasticValue` will set these fields according to value type, and `toQuadValue` will make values with correct types according to fields set in this struct. This should fix these two last tests.

---

Reviewed 1 of 4 files at r8, 1 of 1 files at r11.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 395 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kx9o20UiB8YFLYiS5nl:-Kx9o20UiB8YFLYiS5nm:b-2nn6tk) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L395)):*
> ```Go
> 		}
> 	}
> 	return nil
> ```

You should flush index here:
```golang
_, err := qs.client.Flush(DefaultESIndex).Do(context.TODO())
return err
```

---

*[graph/elastic/quadstore.go, line 582 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kx9oL8wgJh82Bnlkp6S:-Kx9oL8wgJh82Bnlkp6T:bhkxq2g) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L582)):*
> ```Go
> 		Type("nodes").
> 		Query(termQuery).
> 		From(0).Size(1).
> ```

Is `From(0)` necessary?

---

*[graph/elastic/quadstore.go, line 622 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kx9qNmxB1XiOMtbNpw5:-Kx9qNmxB1XiOMtbNpw6:behdo1y) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L622)):*
> ```Go
> 	searchResult, err := qs.client.
> 		Search().
> 		Index(DefaultESIndex).
> ```

Should be `log` index?

---

*[graph/elastic/quadstore.go, line 643 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kx9qIOASjrNDwlJjmqC:-Kx9qIOASjrNDwlJjmqD:bypxc57) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L643)):*
> ```Go
> 	}
> 
> 	eKey, err := strconv.ParseInt(eNode.ID, 10, 64)
> ```

You are parsing hex hash as int. You can use `NewUniqueKey` with IDs generated by Elastic. But not hashes.

---

*[graph/elastic/quadstore_test.go, line 40 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kx9phFoW03zKKaLtX9F:-Kx9phFoW03zKKaLtX9G:b-fh9s9r) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore_test.go#L40)):*
> ```Go
> func TestElasticAll(t *testing.T) {
> 	graphtest.TestAll(t, makeElastic, &graphtest.Config{
> 		SkipNodeDelAfterQuadDel: true,
> ```

Correct config is:
```
{
  NoPrimitives: true,
  SkipIntHorizon: true,
}
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-Kx9s-9-HyeZFp3K-hv1:b-hh690)*

camazotz:
Reviewed 1 of 2 files at r1, 2 of 4 files at r5, 2 of 4 files at r8, 3 of 3 files at r12.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Reviewed 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 322 at r7](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kuahdr2GBm8rZulJyxN:-KxPpx8CR7SrdRycDNEH:b-mtwb4r) ([raw file](https://github.com/cayleygraph/cayley/blob/afd94a7908655faa8b79e32db671d5b9523f1935/graph/elastic/quadstore.go#L322)):*

Thanks for clarification, I forgot about this code path. An approach is totally valid.
There might be performance concerns, but Elastic has pretty good indexing, so it might not be a problem.
</blockquote></details>

@camazotz Please mark this comment as resolved in Reviewable

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Reviewed 2 of 2 files at r13.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-KxPtF9ZmWjEcXe4jgJz:boc9v1f) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*

Let me rerun tests for this one.
</blockquote></details>

Same tests (`TestLoadTypedQuads` and `TestCompareTypedValues`) fail with this error now:
```
elastic: Error 400 (Bad Request): failed to parse [node] [type=mapper_parsing_exception]
```
Which is related to "node" field type set to "string" on index creation. It can be removed, but it leads to this errors:
```
elastic: Error 400 (Bad Request): object mapping for [node] tried to parse field [node] as object,
but found a concrete value [type=mapper_parsing_exception]
```
It seems like Elastic enforces the type of each object field as soon as few values are inserted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Review status: 6 of 7 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-KxUsBBt_ev0kc5ZUqIF:brn823w) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*

Same tests (`TestLoadTypedQuads` and `TestCompareTypedValues`) fail with this error now:
```
elastic: Error 400 (Bad Request): failed to parse [node] [type=mapper_parsing_exception]
```
Which is related to "node" field type set to "string" on index creation. It can be removed, but it leads to this errors:
```
elastic: Error 400 (Bad Request): object mapping for [node] tried to parse field [node] as object,
but found a concrete value [type=mapper_parsing_exception]
```
It seems like Elastic enforces the type of each object field as soon as few values are inserted.
</blockquote></details>

I pushed some changes that I think should fix this. I was able to write using the Cayley write API and run queries on them. Is there a way to run these tests on my end? I tried:
```
go test -run graphtest.go 
?   	github.com/cayleygraph/cayley/graph/graphtest	[no test files]
```
Is there a different syntax to run the test?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Reviewed 1 of 1 files at r14.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-Ky18K5Pmx1bIAwNPqfj:b-1tvkum) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

I pushed some changes that I think should fix this. I was able to write using the Cayley write API and run queries on them. Is there a way to run these tests on my end? I tried:
```
go test -run graphtest.go 
?   	github.com/cayleygraph/cayley/graph/graphtest	[no test files]
```
Is there a different syntax to run the test?
</blockquote></details>

You need Docker to be installed to run these tests. Here is an example of the command:
```
go test -tags docker ./graph/elastic/...
```

---

*[graph/elastic/quadstore.go, line 208 at r14](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Ky18oNgJQmoe8YBmWYm:-Ky18oNgJQmoe8YBmWYn:bm19o4h) ([raw file](https://github.com/cayleygraph/cayley/blob/3a1f51ef5d374d1256f33c6f49a04bb83461b4da/graph/elastic/quadstore.go#L208)):*
> ```Go
> 		return quad.Time(d)
> 	case map[string]interface{}:
> 		if _, ok := d["iri"].(bool); ok {
> ```

Each of these cases extracts `d["val"].(string)`. It's better to move this line to the beginning of the case, and make safe assert:
```golang
val, _ := d["val"].(string)
// if branches
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Reviewed 1 of 1 files at r14.
Review status: 6 of 7 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
@dennwc could you review the changes and let us know if it looks good? Thanks!

dennwc:
Reviewed 1 of 1 files at r15.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-KySP3T4S2twDMLy5JSK:b-f9mg5z) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*

You need Docker to be installed to run these tests. Here is an example of the command:
```
go test -tags docker ./graph/elastic/...
```
</blockquote></details>

Nothing changed here, I still see the same error:

```
elastic: Error 400 (Bad Request): failed to parse [node] [type=mapper_parsing_exception]
```

You could replicate the same result by rebasing changes on top of current `master` branch and by changing test as follows:

```golang
// ...
conf.Image = "elasticsearch"
conf.OpenStdin = true
conf.Tty = true

addr, closer := dock.RunAndWait(t, conf, dock.WaitPort("9200"))
addr = "http://" + addr + ":9200"
// ...
```

An exact command to run tests is:
```bash
go test -v -tags docker ./graph/elastic/... -run TestElasticAll/load_typed_quad
```

As I see it now, Elastic disallows to place multiple types into a single field. An error is because `toElasticValue` may return a different type for each node and writes it to the same field, while Elastic expects it to be the same. Mongo works with this approach, but for Elastic we should do it differently. I would say `node` should always be an object, and it may contain a field per type: `str` or `val_str` for string, `int` for int, etc.

---

*[graph/elastic/quadstore.go, line 250 at r15](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KySNlX3sNAvr417gbQ-:-KySNlX3sNAvr417gbQ0:bi68rft) ([raw file](https://github.com/cayleygraph/cayley/blob/09a43760f1bccfe82a4b101ecb79d84c51e37532/graph/elastic/quadstore.go#L250)):*
> ```Go
> 		return nil
> 	}
> 	fmt.Println("to elastic value-----")
> ```

Remove debug print

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-Kyw-gHyiKL-nyPKcp3A:b-5bep3e) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*

Nothing changed here, I still see the same error:

```
elastic: Error 400 (Bad Request): failed to parse [node] [type=mapper_parsing_exception]
```

You could replicate the same result by rebasing changes on top of current `master` branch and by changing test as follows:

```golang
// ...
conf.Image = "elasticsearch"
conf.OpenStdin = true
conf.Tty = true

addr, closer := dock.RunAndWait(t, conf, dock.WaitPort("9200"))
addr = "http://" + addr + ":9200"
// ...
```

An exact command to run tests is:
```bash
go test -v -tags docker ./graph/elastic/... -run TestElasticAll/load_typed_quad
```

As I see it now, Elastic disallows to place multiple types into a single field. An error is because `toElasticValue` may return a different type for each node and writes it to the same field, while Elastic expects it to be the same. Mongo works with this approach, but for Elastic we should do it differently. I would say `node` should always be an object, and it may contain a field per type: `str` or `val_str` for string, `int` for int, etc.
</blockquote></details>

I installed go-dockerclient but when I run the test, I get this error:
```
--- SKIP: TestElasticAll/load_typed_quad (0.00s)
    	dock.go:35: Post http://unix.sock/containers/create?: proxyconnect tcp: dial unix /var/run/docker.sock: connect: no such file or directory
```
Is there anything else I need to change in the code for this to work?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Reviewed 2 of 4 files at r16.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-Kz3yT2PRwKy2rACewG3:bpivxh8) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

I installed go-dockerclient but when I run the test, I get this error:
```
--- SKIP: TestElasticAll/load_typed_quad (0.00s)
    	dock.go:35: Post http://unix.sock/containers/create?: proxyconnect tcp: dial unix /var/run/docker.sock: connect: no such file or directory
```
Is there anything else I need to change in the code for this to work?
</blockquote></details>

Sorry for delay.

You need to install Docker as well:

```bash
curl -fsSL get.docker.com -o get-docker.sh
sh get-docker.sh
sudo usermod -aG docker $your_user
```

And pull an ElasticSearch image once:

```bash
docker pull elasticsearch
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KzuGXxLblWv9DGBAfLL:-KzuGXxLblWv9DGBAfLM:by4zl96) (no related file):*
There is a second integration option now. I done some work on moving generic parts of NoSQL backend to a separate package, and rewritten Mongo-specific part on top of it. It should be easier to work with, so if you want, you can check this [file](https://github.com/dennwc/cayley/blob/nosql-indexes/graph/nosql/nosql.go)  to see an interface that needs to be implemented. In any case it's okay to finish with current approach and port it later.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Reviewed 2 of 2 files at r17.

---

*[graph/elastic/quadstore.go, line 256 at r17](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KzzYayjogKAeKNlS37g:-KzzYayjogKAeKNlS37h:b-56d0ok) ([raw file](https://github.com/cayleygraph/cayley/blob/b9c2e614185e6a9de14fdc02ca3795f7c8c669dc/graph/elastic/quadstore.go#L256)):*
> ```Go
> 	switch d := v.(type) {
> 	case quad.Raw:
> 		return elasticString{StringVal: string(d)}
> ```

You'll need a flag `IsRaw: true` to distinguish it from String.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:

---

*[graph/elastic/quadstore.go, line 208 at r17](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-Kzz_gac1LrJM2ocLUa5:-Kzz_gac1LrJM2ocLUa6:b-6r29j7) ([raw file](https://github.com/cayleygraph/cayley/blob/b9c2e614185e6a9de14fdc02ca3795f7c8c669dc/graph/elastic/quadstore.go#L208)):*
> ```Go
> 		return quad.Time(d)
> 	case map[string]interface{}:
> 		val, _ := d["val"].(string)
> ```

Now this code needs to decode value differently. It may also decode directly to `elasticString` now, since all fields have a strictly defined type. This should simplify decoding path. Only thing to consider is `[]byte` case that will need a separate `val_bytes` field.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Review status: 4 of 6 files reviewed at latest revision, 4 unresolved discussions.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-KzzmgBg-o1sU51v8UjV:bizl7tu) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*

Sorry for delay.

You need to install Docker as well:

```bash
curl -fsSL get.docker.com -o get-docker.sh
sh get-docker.sh
sudo usermod -aG docker $your_user
```

And pull an ElasticSearch image once:

```bash
docker pull elasticsearch
```
</blockquote></details>

I think I've fixed the type conversion issues. Could you run the tests on your end and let me know if it checks out? I'll set up cayley on a linux VM so I can run the docker tests on my own in the future.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
Reviewed 1 of 2 files at r18, 1 of 1 files at r19.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-6AEgfgywfKPeKgLcm:-L-6AEgfgywfKPeKgLcn:bei3tku) (no related file):*
The last thing we need is to run integration tests - it will run few advanced queries, and might show some hidden issues with iterators. Will run the suit later today.

---

*[graph/elastic/quadstore.go, line 234 at r11](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-KwoLRrYOXM4pSyeRr8c:-L-69VHNLuAQIJWK2vgc:b-n8eo8p) ([raw file](https://github.com/cayleygraph/cayley/blob/9bf4bf2057ee624ae1c24c12ec25e52b9d6ab510/graph/elastic/quadstore.go#L234)):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

I think I've fixed the type conversion issues. Could you run the tests on your end and let me know if it checks out? I'll set up cayley on a linux VM so I can run the docker tests on my own in the future.
</blockquote></details>

Congratulations, all tests are green, assuming all changes above :)
Please mark this thread as resolved.

---

*[graph/elastic/quadstore.go, line 194 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-68S7ybc-CIg2l7mou:-L-68S7zBOMh0ZSVTE-m:b-qku9vc) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L194)):*
> ```Go
> 	val := d.StringVal
> 
> 	if d.IsIRI == true {
> ```

`if d.IsIRI {`, same for BNode

---

*[graph/elastic/quadstore.go, line 194 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-6AlvOriHkdjTOywEF:-L-6AlvOriHkdjTOywEG:bcfl89p) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L194)):*
> ```Go
> 	val := d.StringVal
> 
> 	if d.IsIRI == true {
> ```

You may also use `switch` here:
```golang
switch {
case d.IsIri:
  // ...
case d.IntVal != nil:
  // ...
  // ...
default:
  return quad.String(val)
}
```

---

*[graph/elastic/quadstore.go, line 199 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-6BBLLGtQNqlqEGrrX:-L-6BBLLGtQNqlqEGrrY:b-up28ed) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L199)):*
> ```Go
> 		return quad.BNode(val)
> 	} else if d.IntVal != nil {
> 		val := *d.IntVal
> ```

Inline all the temp vars: `return quad.Int(*d.IntVal)`

---

*[graph/elastic/quadstore.go, line 212 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-68cwaSm22uFHhIm2m:-L-68cwaSm22uFHhIm2n:by5klja) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L212)):*
> ```Go
> 	} else if d.RawVal != nil {
> 		val := *d.RawVal
> 		return quad.String(val)
> ```

`return quad.Raw(val)`

---

*[graph/elastic/quadstore.go, line 236 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-69dUuRb8jDO7zTb_z:-L-69dUuRb8jDO7zTba-:b1wmwln) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L236)):*
> ```Go
> 
> type elasticString struct {
> 	StringVal string     `json:"val_str"`
> ```

Please add `omitempty` back - it will work properly with pointers

---

*[graph/elastic/quadstore.go, line 241 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-68kNRorpWTUI2jIdu:-L-68kNRorpWTUI2jIdv:bhknoy6) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L241)):*
> ```Go
> 	BoolVal   *bool      `json:"val_bool"`
> 	TimeVal   *time.Time `json:"val_time"`
> 	RawVal    *quad.Raw  `json:"val_raw"`
> ```

It's better to set `IsRaw` flag and store value in `StringVal` - this way we can index these values along with any other string-like values

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
:lgtm:

---

Reviewed 1 of 4 files at r16, 1 of 2 files at r17, 1 of 2 files at r18, 1 of 1 files at r19.
Review status: 5 of 6 files reviewed at latest revision, 8 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-L-86hDeQ2R9yjzjl_yU:bnfp4nl)*

dennwc:
Reviewed 1 of 1 files at r20.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-6AEgfgywfKPeKgLcm:-L-cMptzUdbKXvoopKhS:b-u0izkl) (no related file):*

The last thing we need is to run integration tests - it will run few advanced queries, and might show some hidden issues with iterators. Will run the suit later today.
</blockquote></details>

Integration tests are blocked by the fact that write performance is poor because of many small writes. We can batch most writes, but this work should be done in [nosql](https://github.com/dennwc/cayley/tree/nosql) branch.

So I would say we should merge this PR (after last small comment is resolved) and then port it to `nosql` branch as a separate PR. This will allow us to reuse batch code for all `nosql` backends to get performance back, and will simplify testing significantly.

---

*[graph/elastic/quadstore.go, line 236 at r19](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-L-69dUuRb8jDO7zTb_z:-L-8NKLBL1w2OgoK1Izd:b-ldps8n) ([raw file](https://github.com/cayleygraph/cayley/blob/5944606977a00df0520e4c21bb578c6cd9f90b3b/graph/elastic/quadstore.go#L236)):*

Please add `omitempty` back - it will work properly with pointers
</blockquote></details>

and the same for other `val_*` fields

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

camazotz:
Reviewed 1 of 1 files at r20.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634)*

dennwc:
:lgtm:

---

Reviewed 1 of 1 files at r21.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/634#-:-L-dPPW5ucKA3XYUUKzk:bnfp4nl)*

dennwc:
@camazotz @michaelqiu94 Thank you for contributing this large piece of work! :)

The next step, as I mentioned before might be a port to `nosql` meta-backend. It should remove most duplicated code that was shared with Mongo and will leave only parts specific to Elastic.

For now, the potential port will not pass tests because we were not expecting that Elastic can't change value types at runtime. Thus, as a first step, I will change `nosql` implementation to use the same field names as Elastic (a field per value type). After this I believe port should be seamless.

PR #635, opened 2017-09-21
Custom quad metadata support and querying capability

camazotz:
Quadmetadata Proposal -
 https://hackmd.io/BwdgZgJhCmBGwFoBMJpgQFhAQ2g20EAbAmPEgIwDMAnMDSEjUA==

**Salient features in this pull request**
Add a field called 'QuadMetadata' as another direction to quad.Quad.
QuadMetadata is a map[string]string and relies on the Value type 'QuadRef'.
Modify usage of quad getDirection in the Elastic backend to support this field.
Add Gizmo functions InE, OutE, and BothE to support metadata querying
Add time range querying and regex querying capability on quadmetadata to the elastic backend
Pass in elasticsearch index name as an option (through cayley_example.yml file) instead of hardcoding

Elastic implementation details:

- The values in the map are used as unique identifiers for a quad (they are used to form the hash value for the quad). For example, if a given quad has metadata {"index1": "val1", "index2": "val2"}, and another quad has metadata {"index1": "val3", "index2": "val4"}, then these quads are unique.
- QuadMetadata is not a node; it is associated only with the quad.
- OutE function syntax (same for InE and BothE): path.OutE(predicate,{metadata_field: filter_expression})
- Time range query example: g.V("SubjectNode").OutE("PredicateNode", {"field1":"value1", "timestamp":"2015-01-01=>2015-01-15"}).All()
The arrow symbol '=>' is used as a separator between the starting time and ending time. Any field that is set as type 'date' in its elastic mapping will be interpreted as a time range query.
- Regex query example: g.V("SubjectNode").OutE("PredicateNode", {"field1":"item.*", "field2":"v.*", "timestamp":"2015-01-01=>2015-01-15"}).All()
Any field that is not set as type 'date' in its elastic mapping can be queried with a regex expression. In the example above, field1 and field2 are queried with regex.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/635)


dennwc:
This should wait until https://github.com/cayleygraph/cayley/pull/634 is finished and rebased on top of it.
Plus, we will probably have a more generic implementation of quad metadata from our side.

---

Review status: 0 of 12 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-:-Kuap3qmENVjjMT-5zjC:bwn4mfj)*

camazotz:
@dennwc Agreed, #634 comes first. We had a discussion with Barak on the quad metadata implementation and came up with this proposal - https://hackmd.io/BwdgZgJhCmBGwFoBMJpgQFhAQ2g20EAbAmPEgIwDMAnMDSEjUA==

If you have thoughts on a more generic implementation, I'd be happy to take that on.

dennwc:
Reviewed 1 of 4 files at r5, 1 of 11 files at r8, 5 of 14 files at r9.
Review status: 4 of 15 files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PpaJurBL3dnP10hOw:-L0PpaJurBL3dnP10hOx:b-h56995) (no related file):*
The problem with this PR is that in only allows writing metadata to Elastic and nothing else. We should also design a solution for other backends.

---

*[cayley_example.yml, line 10 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0Pq1HT3GG6yHmsBDEk:-L0Pq1HT3GG6yHmsBDEl:b-hx5lpu) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/cayley_example.yml#L10)):*
> ```YAML
>   # backend-specific options
>   options:
>     index: "cayley"
> ```

bolt backend has no `index` option

---

*[data/testdata2.nq, line 1 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PpuuU17OBECoSEgPK:-L0PpuuU17OBECoSEgPL:b-eenyj8) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/data/testdata2.nq#L1)):*
> ```
> <brad> <attime> "2015-01-01" .
> ```

I haven't seen  any tests using this file. Is it necessary?

---

*[quad/quad.go, line 51 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0Pn7AMjkpgtEuy4FB0:-L0Pn7AMjkpgtEuy4FB1:br4j2hq) ([raw file](https://github.com/cayleygraph/cayley/blob/01a19016c2973ef26e47f7ceffac8d3c8ce487e5/quad/quad.go#L51)):*
> ```Go
> 
> // Make creates a quad with provided values.
> func Make(subject, predicate, object, label interface{}, quadmetadata interface{}) (q Quad) {
> ```

This will break compatibility. This should be a separate method `MakeWithMeta` or something similar

---

*[quad/quad.go, line 122 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PnZ2s0_lLuLUFbg5S:-L0PnZ2s0_lLuLUFbg5T:b-idyg7f) ([raw file](https://github.com/cayleygraph/cayley/blob/01a19016c2973ef26e47f7ceffac8d3c8ce487e5/quad/quad.go#L122)):*
> ```Go
> 	Object       Value `json:"object"`
> 	Label        Value `json:"label,omitempty"`
> 	QuadMetadata Value `json:"quadmetadata,omitempty"`
> ```

I guess it would be better to make a field `Metadata map[string]Value` or `Metadata map[quad.Direction]Value` and make `quad.Direction` a string/IRI. The point is - metadata should not be any different from `Label` (we will just leave it as a separate field for compatibility). 

---

*[quad/value.go, line 137 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PooQjEXvdIcdO1w1d:-L0PooQjEXvdIcdO1w1e:b-ogduqc) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/quad/value.go#L137)):*
> ```Go
> 
> // QuadRef is a map from string to any value for holding extra Quad metadata
> type QuadRef map[string]string
> ```

The problem with declaring maps as `Value`s is that user will be able to write quads with this map in any other direction: subject, object, predicate. And I'm not even sure what this should mean in RDF world. So please check comment about making Metadata a map.

---

*[query/gizmo/traversals.go, line 22 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PoQQbI48D_T2pAxrd:-L0PoQQbI48D_T2pAxre:b-pi9d9n) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/query/gizmo/traversals.go#L22)):*
> ```Go
> 	"fmt"
> 
> 	//"github.com/cayleygraph/cayley/vendor/github.com/dop251/goja"
> ```

remove this

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/635)*

camazotz:
Review status: 4 of 15 files reviewed at latest revision, 7 unresolved discussions, some commit checks broke.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PpaJurBL3dnP10hOw:-L0Qle910smqtrQEoRy-:b-in8nue) (no related file):*

The problem with this PR is that in only allows writing metadata to Elastic and nothing else. We should also design a solution for other backends.
</blockquote></details>

Yes, I guess this PR has two parts. First part is altering structure of Quad to support the Metadata field and adding Gizmo methods to query it - this would be applicable across all backends. 
The second part is backend-specific changes to support the Metadata field - even if this part was not added immediately for other backends, the updated Quad structure would not break the existing implementations (they just wouldn't be dealing with the Metadata field for the time being).

---

*[cayley_example.yml, line 10 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0Pq1HT3GG6yHmsBDEk:-L0QoyfqigIS-PbwOrko:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/cayley_example.yml#L10)):*

bolt backend has no `index` option
</blockquote></details>

Done.

---

*[data/testdata2.nq, line 1 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PpuuU17OBECoSEgPK:-L0QlWXNRR5M-9j7WNcX:b8u8g9g) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/data/testdata2.nq#L1)):*

I haven't seen  any tests using this file. Is it necessary?
</blockquote></details>

No it's not, will remove.

---

*[quad/quad.go, line 51 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0Pn7AMjkpgtEuy4FB0:-L0QczpKfMkNqQc1bi0_:b-1qycya) ([raw file](https://github.com/cayleygraph/cayley/blob/01a19016c2973ef26e47f7ceffac8d3c8ce487e5/quad/quad.go#L51)):*

This will break compatibility. This should be a separate method `MakeWithMeta` or something similar
</blockquote></details>

Why does it break compatibility? The metadata case is addressed in an if statement:
```
if q.QuadMetadata, ok = AsValue(quadmetadata); !ok {
		q.QuadMetadata = String(fmt.Sprint(quadmetadata))
	}
```

---

*[quad/quad.go, line 122 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PnZ2s0_lLuLUFbg5S:-L0Qd_ZeuITc03ceXWG9:b-jhrmxx) ([raw file](https://github.com/cayleygraph/cayley/blob/01a19016c2973ef26e47f7ceffac8d3c8ce487e5/quad/quad.go#L122)):*

I guess it would be better to make a field `Metadata map[string]Value` or `Metadata map[quad.Direction]Value` and make `quad.Direction` a string/IRI. The point is - metadata should not be any different from `Label` (we will just leave it as a separate field for compatibility). 
</blockquote></details>

I agree on the name change. My initial thought was to make it type `map[string]Value` but a lot methods in quadstore.go and iterator.go that are used in queries rely on the quad.Value and graph.Value types. Some examples:
```
func (qs *QuadStore) ValueOf(s quad.Value) graph.Value {

func (qs *QuadStore) NameOf(v graph.Value) quad.Value {

func hashOf(s quad.Value) string {

func toQuadValue(d elasticString) quad.Value {

func NewIterator(qs *QuadStore, resultType string, d quad.Direction, val graph.Value) *Iterator {
```
Which is why I thought it would be better to make the entire map as a Value so that I could refer to the entire Metadata field as a Value type while performing queries and not write special cases to parse it.

With regard to `map[quad.Direction]Value`, I thought quad.Direction was meant to refer to the fields of a quad and not any generic String, as below:
```
// Direction specifies an edge's type.
type Direction byte

// List of the valid directions of a quad.
const (
	Any Direction = iota
	Subject
	Predicate
	Object
	Label
	QuadMetadata
)

var Directions = []Direction{Subject, Predicate, Object, Label, QuadMetadata}
```

---

*[quad/value.go, line 137 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PooQjEXvdIcdO1w1d:-L0QkH7X8knjeGCqT4bU:b3waz2k) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/quad/value.go#L137)):*

The problem with declaring maps as `Value`s is that user will be able to write quads with this map in any other direction: subject, object, predicate. And I'm not even sure what this should mean in RDF world. So please check comment about making Metadata a map.
</blockquote></details>

Making Metadata a map causes problems with a lot of the existing methods in quadstore.go and iterator.go that rely on quad.Value type (I gave more detail on the other comment). I don't think the user would be able to write maps in another direction though because the JSON is marshalled into a rawQuad struct, which declares the types explicitly (quad.go):

```
type rawQuad struct {
	Subject      string            `json:"subject"`
	Predicate    string            `json:"predicate"`
	Object       string            `json:"object"`
	Label        string            `json:"label,omitempty"`
	QuadMetadata map[string]string `json:"quadmetadata,omitempty"`
}

func (q Quad) MarshalJSON() ([]byte, error) {
	rq := rawQuad{
		Subject:   q.Subject.String(),
		Predicate: q.Predicate.String(),
		Object:    q.Object.String(),
	}
	if q.Label != nil {
		rq.Label = q.Label.String()
	}
	if q.QuadMetadata != nil {
		rq.QuadMetadata = q.QuadMetadata.Native().(map[string]string)
	}
	return json.Marshal(rq)
}
```
So it would cause a type conversion error if a user tried to write a map into Subject for example.

---

*[query/gizmo/traversals.go, line 22 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PoQQbI48D_T2pAxrd:-L0QpHF4GFDgvjjzfyW9:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/query/gizmo/traversals.go#L22)):*

remove this
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/635)*

dennwc:
Reviewed 2 of 3 files at r10.
Review status: 2 of 13 files reviewed at latest revision, 5 unresolved discussions, some commit checks broke.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0PpaJurBL3dnP10hOw:-L0Xqlpr47E2KF9LOk7L:b15hxno) (no related file):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

Yes, I guess this PR has two parts. First part is altering structure of Quad to support the Metadata field and adding Gizmo methods to query it - this would be applicable across all backends. 
The second part is backend-specific changes to support the Metadata field - even if this part was not added immediately for other backends, the updated Quad structure would not break the existing implementations (they just wouldn't be dealing with the Metadata field for the time being).
</blockquote></details>

I ported Elastic to NoSQL and will soon merge it (#657). Please rebase all your changes on top of it. This will make sure that your changes will work for Mongo as well.

Note: you need to rebase, not merge. In other cases you will have many unrelated changes in this PR.

---

*[graph/shape/path.go, line 14 at r9](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0RK69TLb5ftI5Kh7RH:-L0RK69TLb5ftI5Kh7RI:b-pljak6) ([raw file](https://github.com/cayleygraph/cayley/blob/7f9a4b070d3110f078cc94408f5019ab530c859f/graph/shape/path.go#L14)):*
> ```Go
> 	switch s1 := s1.(type) {
> 	case AllNodes:
> 		fmt.Println("allnodes")
> ```

remove debug prints

---

*[quad/quad.go, line 51 at r8](https://reviewable.io:443/reviews/cayleygraph/cayley/635#-L0Pn7AMjkpgtEuy4FB0:-L0RJnXXTmi23t7tGgD2:b-66wo7a) ([raw file](https://github.com/cayleygraph/cayley/blob/01a19016c2973ef26e47f7ceffac8d3c8ce487e5/quad/quad.go#L51)):*
<details><summary><i>Previously, camazotz (Navalava) wrote…</i></summary><blockquote>

Why does it break compatibility? The metadata case is addressed in an if statement:
```
if q.QuadMetadata, ok = AsValue(quadmetadata); !ok {
		q.QuadMetadata = String(fmt.Sprint(quadmetadata))
	}
```
</blockquote></details>

It break compatibility for existing users that use Make - it will now accept 5 args instead of 4.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/635)*

dennwc:
Closing because of conflicts. Feel free to rebase and resubmit PR.

PR #636, opened 2017-09-26
Allow tag removal.

BlakeMScurr:
Allow tag removal across a whole path. 
Useful if you have some path with a tag, and want to exclude some results from TagResults:

```go
newPath := oldPath.Except(
    oldPath.Has("edge", quad.String("UnwantedEdge"))
    .Untag("someTag"))

it := newPath.BuildIteratorOn(store)
it.Next()
it.TagResults(someMap)
```

Without untag, the expected value of `someMap["someTag"]` may have been overwritten by the value in the excepted path.

### DIFF
--- a/graph/iterator.go
+++ b/graph/iterator.go
@@ -47,6 +47,25 @@ func (t *Tagger) Add(tag ...string) {
 	t.tags = append(t.tags, tag...)
 }
 
+// Remove tags from the iterator
+func (t *Tagger) Remove(tags []string) {
+	newTags := []string{}
+	for _, tag := range t.tags {
+		keep := true
+		for _, rem := range tags {
+			if tag == rem {
+				keep = false
+			}
+		}
+
+		if keep {
+			newTags = append(newTags, tag)
+		}
+	}
+
+	t.tags = newTags
+}
+
 func (t *Tagger) AddFixed(tag string, value Value) {
 	if t.fixedTags == nil {
 		t.fixedTags = make(map[string]Value)
--- a/graph/path/morphism_apply_functions.go
+++ b/graph/path/morphism_apply_functions.go
@@ -138,6 +138,17 @@ func hasReverseMorphism(via interface{}, nodes ...quad.Value) morphism {
 	}
 }
 
+func untagMorphism(tags ...string) morphism {
+	return morphism{
+		Name:     "untag",
+		Reversal: func(ctx *pathContext) (morphism, *pathContext) { return untagMorphism(tags...), ctx },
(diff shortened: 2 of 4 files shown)


dennwc:
Reviewed 4 of 4 files at r1.

---

*[a discussion](https://reviewable.io:443/reviews/cayleygraph/cayley/636#-Kv2ASSmQxZkA6yDkXRH:-Kv2ASSmQxZkA6yDkXRI:b-r4hcyp) (no related file):*
Just wondering, is this change for removing tags left by `Back`?

---

*[graph/iterator.go, line 51 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/636#-Kv263i0zcJx9g28puFo:-Kv263i0zcJx9g28puFp:br5y1yn) ([raw file](https://github.com/cayleygraph/cayley/blob/40cc01d2a1d4e0dc548a43cec39ddd0ef6b49b94/graph/iterator.go#L51)):*
> ```Go
> 
> // Remove tags from the iterator
> func (t *Tagger) Remove(tags []string) {
> ```

Make it variadic like `Add`

---

*[graph/iterator.go, line 52 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/636#-Kv25t1BwdhHVjWXwa3K:-Kv25t1BwdhHVjWXwa3L:bih4n4l) ([raw file](https://github.com/cayleygraph/cayley/blob/40cc01d2a1d4e0dc548a43cec39ddd0ef6b49b94/graph/iterator.go#L52)):*
> ```Go
> // Remove tags from the iterator
> func (t *Tagger) Remove(tags []string) {
> 	newTags := []string{}
> ```

`:= make([]string, 0, len(t.tags))`

---

*[graph/shape/shape.go, line 1086 at r1](https://reviewable.io:443/reviews/cayleygraph/cayley/636#-Kv26nezwZ779nRsS38j:-Kv26nezwZ779nRsS38k:b-ildzbn) ([raw file](https://github.com/cayleygraph/cayley/blob/40cc01d2a1d4e0dc548a43cec39ddd0ef6b49b94/graph/shape/shape.go#L1086)):*
> ```Go
> }
> 
> type UnSave struct {
> ```

I don't think it deserves a separate shape. The reason is, it's not an actual query operation.

I would say it should be a function that receives a Shape tree and scans over it to find any Save shapes and removes tags from them (and even removes Save if it has no tags after an operation).

This can be achieved by calling `Optimize` on the root shape and passing new `untagOptimizer` as a parameter.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/cayleygraph/cayley/636)*

dennwc:
Closing due to inactivity

PR #646, opened 2017-11-29
updated quickstart application command line for correct switches 

johndouglascm:
It looks as if some of the switches (-i -c) were deprecated:

Typically, i got errors like this:

harperdb@harperdb:~/cayley/cayley_v0.6.1_linux_amd64$ ./cayley load -c=cayley.cfg -quads data/testdata.nq --alsologtostdout=true
flag provided but not defined: -c

I tested all, including -quads.

### DIFF
--- a/docs/Quickstart-As-Application.md
+++ b/docs/Quickstart-As-Application.md
@@ -27,14 +27,14 @@ You can set up a full [configuration file](Configuration.md) if you'd prefer, bu
 
 Examples for each backend:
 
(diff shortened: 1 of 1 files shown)


dennwc:
Please check Quickstart-As-Application in [v.0.6.1](https://github.com/cayleygraph/cayley/blob/v0.6.1/docs/Quickstart-As-Application.md) branch. Master currently tracks v0.7 that added flag shortcuts like `-c`, `-a`, etc and upgraded CLI in general.

I will close this, since it was done for an older release.

PR #649, opened 2017-12-07
Getting started doc

gauravtiwari:
Fix quick start command options.


dennwc:
Thanks for this fix, but docs currently tracks master branch ([0.7](https://github.com/cayleygraph/cayley/milestone/8)), not 0.6.1. We should document that master is a development branch, I guess.

gauravtiwari:
Gotcha! Would probably makes sense to highlight it somewhere 👍

Issue #691, opened 2018-02-20
Optimise cockroachdb backend


dsymonds:
**Description**

The [CockroachDB](https://www.cockroachlabs.com) backend for Cayley in `graph/sql/cockroach`, added in #299 is significantly slower than it could be. cockroachdb/cockroach#17108 advises a transformation that according to my tests can speed up bulk insertions by 4-5x.

I'm willing to work at upstreaming my improved code if this general effort is wanted.

dsymonds:
Sorry, forgot to mention the underlying reason: CockroachDB is a distributed database, so is much more sensitive to round-trips than usual SQL backends. The current transactions that Cayley generates are far too round-trip heavy.

dsymonds:
I'm sorry, I'm not sure how to interpret that. @dennwc, are you saying that you're going to take this on?

dennwc:
@dsymonds I'd like to include this patch into next minor release. If you will work on this, it's awesome, but if not - I will fix it just before the release.

dsymonds:
Okay, I'll see if I can handle it this week; if you don't see a pull request from me then you can interpret that as me not having enough time.  ;-)

PR #695, opened 2018-02-21
Optimise CockroachDB backend

dsymonds:
This flattens the CockroachDB driver to be mostly self-contained for insertions instead of leaning on the Postgres driver. I think that's the best option here since the tradeoffs between the two are quite different, and it isn't a large amount of code anyway.

I've run all the docker & integration tests; all pass except for TestCockroach/qs/integration, which fails both before and after this change (with `transaction is too large to commit: 100005 intents`) and I don't believe is affected by this change.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -71,7 +71,7 @@ Which will also resolve the relevant static content paths for serving HTTP.
 
 # Running Unit Tests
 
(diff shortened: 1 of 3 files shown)


dsymonds:
@dennwc This is for #691.

dennwc:
Thanks for taking your time to work on this!

PR #697, opened 2018-02-26
Switch CockroachDB driver from lib/pq to jackc/pgx.

dsymonds:
This was at the behest of CockroachDB engineers who say this is a much
better driver
(https://github.com/cockroachdb/cockroach/issues/17108#issuecomment-368024371).

I'm not sure if this is the kind of thing you're interested in accepting, but it was quite easy to do, and seems to work fine for me in a real situation.

This extends work done in #691.


barakmich:
Prima facie, I'm into it!

PR #707, opened 2018-04-17
hello_schema: replace TempFile with TempDir

mguentner:
If TempFile is used cayley.NewGraph complains about
tmpfile.Name() not being a dir.

### DIFF
--- a/examples/hello_schema/main.go
+++ b/examples/hello_schema/main.go
@@ -57,16 +57,17 @@ func main() {
 	}
 
 	// File for your new BoltDB. Use path to regular file and not temporary in the real world
-	tmpfile, err := ioutil.TempFile("", "example")
+	tmpdir, err := ioutil.TempDir("", "example")
 	checkErr(err)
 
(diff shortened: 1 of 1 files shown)


examples/hello_schema/main.go
```diff
@@ -57,16 +57,17 @@ func main() {
 	}
 
 	// File for your new BoltDB. Use path to regular file and not temporary in the real world
-	tmpfile, err := ioutil.TempFile("", "example")
+	tmpdir, err := ioutil.TempDir("", "example")
 	checkErr(err)
 
-	defer os.Remove(tmpfile.Name()) // clean up
+	defer os.Remove(tmpdir) // clean up
```
dennwc: Should probably be `RemoveAll`

mguentner: Fixed

mguentner:
For the record, `hello_bolt` appears to require the same fix

dennwc:
Seems so. It would be great if you could send the change for it as well :)

mguentner:
#708

PR #709, opened 2018-04-18
Fix Dockerfile

skydan:
I've spotted that both `dep`'s steps are broken after [recent migration from glide to dep](https://github.com/cayleygraph/cayley/commit/4c752574fe6a139d804fb22e63892b9a1e2c294b#diff-3254677a7917c6c01f55212f86c57fbf).


dennwc:
Thanks for the fix!

Issue #722, opened 2018-06-21
Mongo backend didn't use the correct database name


lynic:
**Description**
I'm trying to use mongodb as backend, and expected to use "cayley" as database name, but instead, the created database is "test".
My cayley.yml looks like following:
```
store:
  # backend to use
  backend: mongo
  # address or path for the database
  address: "user:pass@mongo_host:27017"
  # open database in read-only mode
  read_only: false
  # backend-specific options
  options:
    # nosync: false
    database_name: cayley
query:
  timeout: 30s
load:
  ignore_duplicates: false
  ignore_missing: false
  batch: 10000
```

**Steps to reproduce the issue:**
1. Use above cayley.yml to start cayley in container
2.
3.

**Received results:**
Database "test" was created instead of "cayley".

**Expected results:**
database "cayley" was created.

**Output of `cayley version` or commit hash:**

```
(paste your output here)
```

**Environment details:**

Backend database: `(database and version)`
mongodb

** Root Cause **
At https://github.com/cayleygraph/cayley/blob/v0.7.4/graph/nosql/mongo/mongo.go#L68 , it pass an empty string as db name, and by default, mgo will use "test" as default name.

PR #723, opened 2018-06-21
Correct db name using mongo backend

lynic:
Close: #722


graph/nosql/mongo/mongo.go
```diff
@@ -64,8 +64,12 @@ func dialDB(addr string, opt graph.Options) (*DB, error) {
 	if err != nil {
 		return nil, err
 	}
+	dbName, err := options.StringKey("database_name", nosql.DefaultDBName)
```
dennwc: Since setting the database name in `dialMongo`does not change it, could you please remove it from there?

PR #725, opened 2018-06-29
Added unsigned integer types

jtorvald:


quad/value.go
```diff
@@ -344,6 +358,22 @@ func (s Int) TypedString() TypedString {
 	}
 }
 
+// Uint is a native wrapper for uint64 type.
+//
+// It uses NQuad notation similar to TypedString.
+type Uint uint64
```
dennwc: It's not as simple as adding it into this file :) It should also be handled in protobuf and all backends.

So instead of adding this type, I propose to convert everything to `Int` for now. It will break large `uint64` values, but the fix on the client side will be just `uint64(v.Native().(int64))`.

Later we could consider adding `Uint`, but this will require a more involved PR.

PR #728, opened 2018-07-10
Begin the Glossary

tooolbox:
In coming to start using Cayley, I discovered that the glossary was never begun!

So I put together some definitions I found in my own readings, so that it may be easier for the next person.  I may add to this list in the near future, but I wanted to get the ball rolling.  Anyone who objects to a particular definition is free to edit it.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/728)


tooolbox:
Fantastic!  🎉

PR #733, opened 2018-08-25
Fix Config.idFor() recursive ID lookup in anonymous fields (fixes #732)

### LINKED ISSUE #732: Config.idFor() doesn't correctly handle multiple levels of nested anonymous structs
**Description**

Non-nil ID detection for @id tagged structure not working correctly with nested anonymous structs with more than one level

I have identified the bug and will submit a pull request with a fix

**Steps to reproduce the issue:**
```
type struct A {
  ID quad.IRI `quad:"@id"`
}

type struct B {
  A
}

type struct C {
  B
}

v := C{}
v.ID = quad.IRI("something") // Not-nil

rules, err := c.rulesFor(rt)
...
id, err := c.idFor(rules, rt, rv, "")
// id == nil
```

**Received results:**
See description

**Expected results:**
id = "<something>"

**Output of `cayley version` or commit hash:**

```
243214c801c26d9b36436ed3d92d4ed18298b309
```

**Environment details:**
n/a

phyrwork:
Prefix was not being propagated in second pass so was only correct at
first level where pref == ""


dennwc:
Thanks for addressing this issue! Can you please add a test case as well?

PR #736, opened 2018-09-19
fix in run go test

gzm1997:


dennwc:
Thanks for the change! Can you please rebase on top of the new master? It should fix CI.

PR #741, opened 2018-10-18
Implement Resolver iterator

### LINKED ISSUE #663: Implement Resolver iterator
There might be cases when user wants to run query from a specific set of values. Currently he needs to either use `shape.Lookup` and let optimizer to resolve nodes in batch mode, or resolve values one by one and add them to `iterator.Fixed`. Better way will be to implement a `Resolver` iterator which takes a set of values (`quad.Value`) and will resolve them to `graph.Value` in process of iteration.

phyrwork:
Implement an iterator that takes a slice of nodes and an
associated QuadStore and resolves to their respective
values during iteration.

Resolves #663


graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+	err    error
+	result graph.Value
+}
+
+// Creates a new Resolver iterator.
+func NewResolver(qs graph.QuadStore, nodes ...quad.Value) *Resolver {
+	it := &Resolver{
+		uid:   NextUID(),
+		qs:    qs,
+		nodes: make([]quad.Value, 0, 20),
```
dennwc:
```suggestion
		nodes: make([]quad.Value, 0, len(nodes)),
```

graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	// Enforce uniqueness
+	unique := make(map[quad.Value]bool, 20)
```
dennwc: I would actually expect it to iterate over duplicates. It might resolve duplicates once, but in general, it should preserve an order and the number of values.

graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+	out.tags.CopyFrom(it)
+	return out
+}
+
+func (it *Resolver) String() string {
+	return fmt.Sprintf("Resolver(%v)", it.nodes)
+}
+
+// Register this iterator as a Resolver iterator.
+func (it *Resolver) Type() graph.Type { return graph.Fixed }
```
dennwc:
```suggestion
func (it *Resolver) Type() graph.Type { return graph.Resolver }
```

graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Check if the passed value is equal to one of the nodes stored in the iterator.
+func (it *Resolver) Contains(ctx context.Context, value graph.Value) bool {
+	graph.ContainsLogIn(it, value)
+	for _, node := range it.nodes {
+		if it.qs.ValueOf(node) == value {
```
dennwc: It might be a good idea to cache those values, since `Contains` will be called multiple times.

graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+
+// Next advances the iterator.
+func (it *Resolver) Next(ctx context.Context) bool {
+	graph.NextLogIn(it)
+	if it.index == len(it.nodes) {
+		it.result = nil
+		return graph.NextLogOut(it, false)
+	}
+	node := it.nodes[it.index]
+	value := it.qs.ValueOf(node)
```
dennwc: It's correct to resolve values this way, but it's usually pretty slow.

As an improvement, you can batch value resolution using [this] function. It will check if QuadStore supports batch operations, and if not, will fallback to resolving them one by one.

graph/iterator/resolver.go
```diff
@@ -0,0 +1,161 @@
...
+// Size is the number of m stored.
+func (it *Resolver) Size() (int64, bool) {
+	return int64(len(it.nodes)), true
+}
+
+func (it *Resolver) Stats() graph.IteratorStats {
+	s, exact := it.Size()
+	return graph.IteratorStats{
+		// Lookup cost is size of set
+		ContainsCost: s - int64(it.index),
```
dennwc:
```suggestion
		ContainsCost: s,
```
The assumption is that `Stats` is stateless.

graph/iterator/resolver_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/cayleygraph/cayley"
+	"github.com/cayleygraph/cayley/quad"
+	"context"
+	"github.com/cayleygraph/cayley/graph"
+	"github.com/cayleygraph/cayley/graph/iterator"
+)
+
+func TestResolverIteratorIterate(t *testing.T) {
+	var ctx context.Context
+	qs, err := cayley.NewMemoryGraph()
```
dennwc: Requiring the top-level `cayley` package from low-level `iterators` package is a bit an overkill. Please check other iterator tests, there should be a mock store for such tests.

graph/iterator/resolver_test.go
```diff
@@ -0,0 +1,130 @@
+package iterator_test
+
+import (
+	"testing"
```
dennwc: running `goimports` over it will be nice (analog of `gofmt`)

graph/iterator/resolver_test.go
```diff
@@ -0,0 +1,130 @@
...
+	nodes := []quad.Value{
+		quad.String("1"),
+		quad.String("2"),
+		quad.String("3"),
+		quad.String("4"),
+		quad.String("5"),
+	}
+	expected := make(map[quad.Value]graph.Value)
+	for _, node := range nodes {
+		qs.AddQuad(quad.Make(quad.String("0"), nil, node, nil))
```
dennwc: This only works because it's an in-memory store. It's better to specify a predicate here.

graph/iterator/resolver_test.go
```diff
@@ -0,0 +1,130 @@
...
+	qs, err := cayley.NewMemoryGraph()
+	if err != nil {
+		t.Fatalf("error creating graph: %v", err)
+	}
+	nodes := []quad.Value{
+		quad.String("1"),
+		quad.String("2"),
+		quad.String("3"),
+		quad.String("4"),
+		quad.String("5"),
```
dennwc: Add a duplicate value as well.

graph/iterator/resolver.go
phyrwork: The BatchQuadStore interface looks iffy...
```
type BatchQuadStore interface {
	ValuesOf(ctx context.Context, vals []Value) ([]quad.Value, error)
}
```
Shouldn't ValuesOf take []quad.Value and return []Value?

i.e.

Shouldn't this be named NamesOf?

Regardless the []quad.Value -> []graph.Value interface/function doesn't seem to exist.

I can add this beforehand in another PR.

phyrwork: I will call the []quad.Value -> []graph.Value NamesOf for consistency even though it's backwards to prevent a breaking change.

graph/iterator/resolver.go
```diff
@@ -0,0 +1,173 @@
...
+func (it *Resolver) Tagger() *graph.Tagger {
+	return &it.tags
+}
+
+func (it *Resolver) TagResults(dst map[string]graph.Value) {
+	it.tags.TagResult(dst, it.Result())
+}
+
+func (it *Resolver) Clone() graph.Iterator {
+	out := &Resolver{
```
dennwc: I would rather call `NewResolver`

graph/iterator/resolver.go
```diff
@@ -0,0 +1,173 @@
...
+
+// A Resolver iterator consists of it's order, an index (where it is in the,
+// process of iterating) and a store to resolve values from.
+type Resolver struct {
+	qs     graph.QuadStore
+	uid    uint64
+	tags   graph.Tagger
+	order  []quad.Value
+	values map[quad.Value]graph.Value
+	nodes  map[graph.Value]quad.Value
```
dennwc: It's not safe to use `graph.Value` as a key. See `Key()` method on `graph.Value`.

graph/iterator/resolver.go
phyrwork: And re-build the cache from scratch?

graph/iterator/resolver.go
phyrwork: Whoops.

graph/iterator/resolver.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	it.cached = true
+	return nil
+}
+
+// Check if the passed value is equal to one of the order stored in the iterator.
+func (it *Resolver) Contains(ctx context.Context, value graph.Value) bool {
+	graph.ContainsLogIn(it, value)
+	if !it.cached {
+		it.resolve(ctx)
```
dennwc: It should set `it.err`, same for `Next`

graph/iterator/resolver.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	_, ok := it.nodes[value.Key()]
+	return graph.ContainsLogOut(it, value, ok)
+}
+
+// Next advances the iterator.
+func (it *Resolver) Next(ctx context.Context) bool {
+	graph.NextLogIn(it)
+	if it.index == len(it.order) {
```
dennwc: `>=`, just to be on the safe side

dennwc:
@phyrwork There are only a few small changes left. Do you have time to finish it? Will be happy to merge it.

PR #742, opened 2018-10-18
Add NamesOf to the BatchQuadStore interface

phyrwork:
Requirement exposed by https://github.com/cayleygraph/cayley/pull/741

Add NamesOf to the BatchQuadStore interface.

Add notes that NamesOf and ValuesOf have their names the wrong
way around.

Add a dummy NamesOf to kv/QuadStore to avoid breaking it's
BatchQuadStore implementation.

### DIFF
--- a/graph/kv/quadstore.go
+++ b/graph/kv/quadstore.go
@@ -272,6 +272,25 @@ func (qs *QuadStore) ValuesOf(ctx context.Context, vals []graph.Value) ([]quad.V
 	}
 	return out, last
 }
+
+func (qs *QuadStore) RefsOf(ctx context.Context, nodes []quad.Value) ([]graph.Value, error) {
+	values := make([]graph.Value, len(nodes))
+	err := View(qs.db, func(tx BucketTx) error {
+		for i, node := range nodes {
+			value, err := qs.resolveQuadValue(ctx, tx, node)
+			if err != nil {
+				return err
+			}
+			values[i] = Int64Value(value)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return values, nil
+}
+
 func (qs *QuadStore) NameOf(v graph.Value) quad.Value {
 	ctx := context.TODO()
 	vals, err := qs.ValuesOf(ctx, []graph.Value{v})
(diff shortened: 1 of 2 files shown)


graph/quadstore.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 type BatchQuadStore interface {
-	ValuesOf(ctx context.Context, vals []Value) ([]quad.Value, error)
+	ValuesOf(ctx context.Context, vals []Value) ([]quad.Value, error) // TODO: Should be called NamesOf()
+	NamesOf(ctx context.Context, nodes []quad.Value) ([]Value, error) // TODO: Should be called ValuesOf()
```
dennwc:
```suggestion
	RefsOf(ctx context.Context, nodes []quad.Value) ([]Value, error) // TODO: Should be called ValuesOf()
```
And the `ValuesOf` above is fine. This is a legacy from an old version of Cayley. `NameOf` and `ValueOf` will be removed in v0.8.

graph/kv/quadstore.go
```diff
@@ -272,6 +272,16 @@ func (qs *QuadStore) ValuesOf(ctx context.Context, vals []graph.Value) ([]quad.V
 	}
 	return out, last
 }
+
+// TODO: Proper BatchQuadStore support for this function
```
dennwc: May I ask you to implement it at least partially? :)

Check the `ValueOf`, an algorithm is quite simple: open `View` once, iterate over values inside, call `resolveQuadValue` for each and store the result. But instead of dropping an error as `ValueOf` does, return it here.

graph/quadstore.go
```diff
@@ -45,6 +46,17 @@ func ValuesOf(ctx context.Context, qs QuadStore, vals []Value) ([]quad.Value, er
 	return out, nil
 }
 
+func NamesOf(ctx context.Context, qs QuadStore, nodes []quad.Value) ([]Value, error) {
```
dennwc:
```suggestion
func RefsOf(ctx context.Context, qs QuadStore, nodes []quad.Value) ([]Value, error) {
```

PR #751, opened 2018-12-09
Add sqlite backend

BrendanBall:
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/751)


BrendanBall:
Thanks for the awesome response time :D . I've just pushed some docs. Let me know if you'd like to add anything else.

dennwc:
@BrendanBall That might explain why they fail for the same SQL implementation that works for PG and MySQL.

I was confused by `Unexpected number of results, got:0 expect:2` without any errors, but that's probably a problem with context error propagation in iterators.

If setting a higher timeout helps, I will consider this solved.

Sure, we might need some optimizations for SQLite to be faster for those kinds of queries, but it's worth merging in the current state as well.

BrendanBall:
the test `TestSqlite/qs/integration/three_huge_sets_with_small_intersection` didn't even complete within an hour :/ . So much for quickly adding a sqlite backend :P

dennwc:
OK, I would say that the fact that backend is slow on integration test shouldn't block this PR. We can make some optimizations later, if necessary.

I will check the PR again and merge it if there are no objections.

PR #760, opened 2019-01-07
pull for issue 758, label context ignored for saveOpt

3pCode:
I also noticed saveR, had the same issue so I put a test case and fix for that as well.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/760)


query/gizmo/gizmo_test.go
```diff
@@ -736,3 +736,70 @@ func issue718Nodes() []string {
...
+			g.V("smart_person").LabelContext("<other_graph>").SaveR("<status>", "who").All()
+		`,
+		tag:    "who",
+		expect: []string{"<fred>"},
+	},
+	
+	
+}
+
+func TestGizmoMultiGraph(t *testing.T) {
```
dennwc: Maybe instead of adding a new test we can define a new `file` field in the existing `TestGizmo`?

3pCode: Challenge with that would be the current test method loads the graph upfront, then executes the tests. What I could do is add the file attribute to the test, then change the existing test method to iterate over the two test data sets, and on each iteration only the tests matching the currently graph is executed. That would allow reuse of that test method. I'll update the pull with that

dennwc: The second options might be to have a `map[string][]quad.Quad` and cache all loaded files there. This way we can use any files without affecting the way how tests are executed.

3pCode: Pull updated, take a look and see what you think, essentially similar to the map suggestion, but in two variables currently.

3pCode:
Syntax changes made.

PR #762, opened 2019-02-08
add ability to specify connection pool details for sql connectors

3pCode:
add ability to specify connection pool details for sql connectors. The three properties represent the three configuration settings available on the go sql connection:

maxopenconnections
maxidleconnections
connmaxlifetime

By default the go defaults, are more appropriate for development, specifically the go defaults allow unlimited number of open connections, only 2 idle connections, and a connmaxlifetime of 0 which closes the connection immediately when closed.  For queries which request in a number of calls to the database, this can be costly, especially in the postgres case, which forks a native process for each connection.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/762)


3pCode:
> > Reviewed 2 of 5 files at r4.
> > _[Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/762)_ status: all files reviewed, 1 unresolved discussion (waiting on @3pCode)
> 
> _[graph/sql/quadstore.go, line 130 at r4](https://reviewable.io/reviews/cayleygraph/cayley/762#-LYE1zGi9LA2ZLyv745S:-LYE1zGi9LA2ZLyv745T:bb74lw8) ([raw file](https://github.com/cayleygraph/cayley/blob/ad97ce1edbd26357dc59233fb4b7685351197f7f/graph/sql/quadstore.go#L130)):_
> 
> > ```go
> > 	if err != nil {
> > 		clog.Errorf("Could not retrieve maxopenconnections from options");
> > 		return nil, err
> > ```
> 
> It may be better to wrap an error with `fmt.Errorf` instead of logging it.

Updated to wrap with fmt.Errorf

PR #766, opened 2019-02-14
Enable native byte array fields on structs

fungl164:
Auto-converts []byte arrays to strings behind the scenes so the bytes are not stored as individual integers on the graph.

Examples:
```Go
// Create some struct with native []byte array field insid
type MyStruct struct {
	ID   string `json:"id" quad:"@id"`
	Name string `json:"name"`  // Sample text field
	Data []byte `json:"data"`  // Sample binary field
}

// Register schema
schema.RegisterType(quad.IRI("mystruct"), MyStruct{})

// Write struct to graph
m := &MyStruct{
	ID:   "1",
	Name: "Sample struct with embedded byte array",
	Data: []byte{'H', 'e', 'l', 'l', 'o', ' ', 'W', 'o', 'r', 'l', 'd', '!'},
}
schema.WriteAsQuads(qw, m)


// Write raw bytes to graph
bytes := quad.ByteArr(data.Data)
schema.WriteAsQuads(qw, bytes)
```

### DIFF
--- a/quad/value.go
+++ b/quad/value.go
@@ -91,6 +91,8 @@ func AsValue(v interface{}) (out Value, ok bool) {
 	switch v := v.(type) {
 	case Value:
 		out = v
+	case []byte:
+		out = ByteArr(v)
 	case string:
 		out = String(v)
 	case int:
@@ -268,10 +270,11 @@ const (
 
 // TODO(dennwc): make these configurable
 const (
-	defaultIntType   IRI = schema.Integer
-	defaultFloatType IRI = schema.Float
-	defaultBoolType  IRI = schema.Boolean
-	defaultTimeType  IRI = schema.DateTime
+	defaultIntType     IRI = schema.Integer
+	defaultFloatType   IRI = schema.Float
+	defaultBoolType    IRI = schema.Boolean
+	defaultTimeType    IRI = schema.DateTime
+	defaultByteArrType IRI = schema.ByteArr
 )
 
 func init() {
@@ -288,6 +291,9 @@ func init() {
 	// time types
 	RegisterStringConversion(defaultTimeType, stringToTime)
 	RegisterStringConversion(nsXSD+`dateTime`, stringToTime)
+	// []byte types
+	RegisterStringConversion(defaultByteArrType, stringToByteArr)
+	RegisterStringConversion(nsXSD+`bytes`, stringToByteArr)
 }
 
 var knownConversions = make(map[IRI]StringConversion)
@@ -340,6 +346,10 @@ func stringToTime(s string) (Value, error) {
 		return nil, err
 	}
 	return Time(v), nil
+}
+
+func stringToByteArr(s string) (Value, error) {
+	return ByteArr([]byte(s)), nil
 }
 
 // Int is a native wrapper for int64 type.
@@ -420,6 +430,28 @@ func (s Time) TypedString() TypedString {
 		// TODO(dennwc): this is used to compute hash, thus we might want to include nanos
 		Value: String(time.Time(s).UTC().Format(time.RFC3339)),
 		Type:  defaultTimeType,
+	}
+}
+
+// ByteArr is representation of []byte as a value
+type ByteArr string
+
+func (b ByteArr) String() string {
+	return b.TypedString().String()
+}
+func (b ByteArr) Native() interface{} { return []byte(b) }
+func (b ByteArr) Equal(v Value) bool {
+	t, ok := v.(ByteArr)
+	if !ok {
+		return false
+	}
+	return b == t
+}
+func (b ByteArr) TypedString() TypedString {
+	return TypedString{
+		// TODO(dennwc): this is used to compute hash
+		Value: String(b),
+		Type:  defaultByteArrType,
 	}
 }
 
--- a/schema/writer.go
+++ b/schema/writer.go
@@ -92,7 +92,7 @@ func (w *writer) writeValueAs(id quad.Value, rv reflect.Value, pref string, rule
 				return err
 			}
 		case saveRule:
-			if f.Type.Kind() == reflect.Slice {
+			if f.Type.Kind() == reflect.Slice && f.Type != reflect.TypeOf([]byte(nil)) {
 				sl := rv.Field(i)
 				for j := 0; j < sl.Len(); j++ {
 					if err := w.writeOneValReflect(id, r.Pred, sl.Index(j), r.Rev); err != nil {
--- a/voc/schema/schema.go
+++ b/voc/schema/schema.go
@@ -25,6 +25,8 @@ const (
 	True = Prefix + `True`
 	// Data type: Text.
 	Text = Prefix + `Text`
+	// Data type: ByteArr.
+	ByteArr = Prefix + "ByteArr"
 	// Data type: URL.
 	URL = Prefix + `URL`
 	// Data type: Number.


fungl164:
___
*[quad/value.go, line 453 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYiAjJZ-jwJwtBwDa7b:-LYiNb4N3tAcTseHj8q0:bt0sxuu) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/quad/value.go#L453)):*

To be compatible with RDF we will probably need to encode it differently.

This may act as a reference, although I'm not sure it's the right one:
https://www.w3.org/TR/Content-in-RDF10/#bytesProperty 

So maybe the "canonical RDF encoding" is the base64  with the type mentioned in that section.
</blockquote></details>

I thought about that.  I implemented base64 encoding/decoding.  there is teh https://www.w3.org/TR/xmlschema-2/#base64Binary definition we can use.  How should I define it in the code?

fungl164:
___
*[schema/writer.go, line 95 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYi81g4Adr8MEzd7CxI:-LYiQH7N1Aufoe9WhYlS:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/schema/writer.go#L95)):*

It's better to save the result of `reflect.TypeOf([]byte(nil))` to a global and use it here
</blockquote></details>

Done.

fungl164:
___
*[voc/schema/schema.go, line 29 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYi8JP35clf65teMsI7:-LYiQuZL30-nmcxMsg7d:b35ede9) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/voc/schema/schema.go#L29)):*

This file corresponds to the types defined by https://schema.org, and there is no such type as "ByteArr".
</blockquote></details>

I just made it up so I could test and run the code.  I don't think there is a definition for just plain byte array data.  Will it break things if we just take it out?

fungl164:
For the tests, I might need your help with the protobuf marshalling/unmarshalling. 

With Bolt, I see the marshalling of quad.Bytes to ValueRaw going through, but the return ToNative() returns an empty array so the test fails.

Here's what I have so far, but you may be faster at figuring out where/what's failing... : )

```Go
func MakeValue(qv quad.Value) *Value {
	...
	case quad.Bytes:
		return &Value{&Value_Raw{v.Native().([]byte)}}
	...
}


// ToNative converts protobuf Value to quad.Value.
func (m *Value) ToNative() (qv quad.Value) {
	...
	case *Value_Raw:
		return quad.Bytes(v.Raw)
	...
}
```

fungl164:
Inching along...Defined ```quad.Bytes``` as proper ```[]byte``` wrapper and I see the bytes getting stored as typed strings: ```"Hello World!"^^<schema:Bytes>``` on  both Badger and Bolt.

The ```LoadTypedQuad``` Test is still not passing even though it's stored.  I suspect resolving the bucket key is part of the issue, but I don't quite understand how the keying and indexing work yet...any comments/pointers/help very much appreciated...  

Thnxs!

fungl164:
___
*[internal/mapset/map.go, line 36 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0K09j7d8jA6EEb9mP:-LZ3Nq2MEGuSQir9EVwI:b-88p6pl) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/internal/mapset/map.go#L36)):*

Do we really need a B-Tree for this simple use case? Why Go maps cannot be used?
</blockquote></details>

Go maps cough up when using quad.Bytes as a map key since it is defined by a slice of bytes.  Btree is just a choice.  It can be replaced with any proper map-like struct.  Perhaps we can do some background trickery to convert strings <-> []bytes in the background...

fungl164:
___
*[internal/mapset/set.go, line 7 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0KJ2Y1TK2iRtKpr6I:-LZ3Rb4A4ac6TeaxcOT0:b-cmkbjr) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/internal/mapset/set.go#L7)):*

Same here, do we really need a generic `Set` interface? Why not define a specific implementation?
</blockquote></details>

Not used. will take it out.

fungl164:
___
*[quad/value.go, line 450 at r6](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZF3cmPEZYq710cioyO:-LZFsMWZ9oBuOp0oqWZy:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/e5da3c5568b052860399155dfbbafee97e65b47d/quad/value.go#L450)):*

It should store raw bytes in the value, not base64. It should convert to base64 only in `TypedValue`.
</blockquote></details>

Done.

fungl164:
___
*[quad/pquads/quads.go, line 23 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0PBkQ5BFHHZZr1f6r:-LZFsssM1oW51bnhO5Ac:bqqhu2e) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/quad/pquads/quads.go#L23)):*

It may be better to remove this case and add a generic one before `default`. This case may check for the `TypedString()` method on the value (define an interface somewhere) and convert it.

So if it recognizes a specific type - it converts directly. If it doesn't, but the values can be converted to `TypeString` - it will be encoded this way as a fallback.
</blockquote></details>

Fixed to store as raw bytes.

fungl164:
___
*[quad/pquads/quads.go, line 97 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0Pu1D0rvwYG448vJX:-LZFt9X2D6wwEHttPH9N:b-6otj1t) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/quad/pquads/quads.go#L97)):*

Instead of asserting for a specific type, there should be references to `TypedString.ParseValue()`call - it will check for the known value types and will parse the string with a specified function. See `quad.RegisterStringConversion`.
</blockquote></details>

Fixed. Removed.

fungl164:
___
*[voc/schema/schema.go, line 29 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYi8JP35clf65teMsI7:-LZFtrI46rb9HecK1Bes:b-d4ywp5) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/voc/schema/schema.go#L29)):*

Yeah, it should be removed from here, but can be added as an unexported constant to `quad` package.
</blockquote></details>

If we removed this, what does the TypedString type become?

fungl164:
___
*[internal/mapset/comparator.go, line 52 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0JffxE53ubZSwkEA2:-LZFu15JAPwa4wfIDAcf:bss4a2b) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/internal/mapset/comparator.go#L52)):*

This case can be used for every other type except `Time` and `[]byte`.
</blockquote></details>

Removed.

fungl164:
___
*[internal/mapset/map.go, line 36 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0K09j7d8jA6EEb9mP:-LZFu4vHD2SjY_SH0qIO:bss4a2b) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/internal/mapset/map.go#L36)):*

They have a special optimization for bytes if you use it like this:
```golang
m := make(map[string]interface{})
// cast to string should be inline for optimization to work
m[string(b)] = x
// this won't be optimized
k := string(b)
m[s] = x
```
</blockquote></details>

Removed.

fungl164:
___
*[internal/mapset/set.go, line 7 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0KJ2Y1TK2iRtKpr6I:-LZFu8kl15tbeCjx6x4b:bss4a2b) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/internal/mapset/set.go#L7)):*
<details><summary><i>Previously, fungl164 wrote…</i></summary><blockquote>

Not used. will take it out.
</blockquote></details>

Removed.

fungl164:
___
*[quad/value.go, line 453 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYiAjJZ-jwJwtBwDa7b:-LZG3BLn0JfKIUMZ9heL:b26a67l) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/quad/value.go#L453)):*

There should be some canonical URI/IRI for this type. I'm not sure `https://www.w3.org/TR/xmlschema-2/#` is the right namespace, though. It may be somewhere in XML namespaces - RDF inherits some of them.
</blockquote></details>

I'm thinking converting raw bytes to b64 and wrapping the result as CDATA would actually be RDF-compliant.  This may actually be a workable solution since CDATA makes no guarantees and leaves it up to the reader to interpret the contents inside the tag...Just need to ensure the docs reflect what the interpretation should be...or we could just skip the CDATA encapsulation and just rely on the docs...what do you think?

fungl164:
___
*[quad/value.go, line 453 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYiAjJZ-jwJwtBwDa7b:-LZGHIkO8_WPQ9kC8Cdj:bw6j8y6) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/quad/value.go#L453)):*

CDATA only works for XML form of RDF, which is rarely used. So we will need a namespace anyway for it to work with NQuads/Turtle/JSON-LD formats.
</blockquote></details>

Will need you to take the lead on this.

fungl164:
___
*[voc/schema/schema.go, line 29 at r1](https://reviewable.io/reviews/cayleygraph/cayley/766#-LYi8JP35clf65teMsI7:-LZGHxtd3PmTKzJfCLS5:b-gt1lf6) ([raw file](https://github.com/cayleygraph/cayley/blob/8b2ad736a6ff7a835fee0ebf93f4cd3fcabe1798/voc/schema/schema.go#L29)):*

We just need to find the right type, and it can be removed from here (since it's not related to schema.org), and added as a constant directly in `quad` package.
</blockquote></details>

Ok. I'll remove it when we find the right type.

fungl164:
___
*[graph/graphtest/graphtest.go, line 710 at r7](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZGAjzsEfcOqy8fRtLg:-LZGHEvxBShaKIXZxsHI:b-3tk5pr) ([raw file](https://github.com/cayleygraph/cayley/blob/84b79193a6ed422cf624b495c6d5f62bf8fb8d64/graph/graphtest/graphtest.go#L710)):*

Sorry, I meant `quad.IRI("C")` -> `quad.IRI("bytes")`. And `quad.Raw("<bytes>")` -> `quad.Bytes("<bytes>")`. This should create the collision that I was talking about.
</blockquote></details>

I think I see what you meant.  I faked-forced a different hash for Bytes vs IRI by inserting an extra byte.  Dunno if its the best way or if there will be other clashes, but hopefully unlikely if the hash is strong enough... : )

fungl164:
___
*[graph/graphtest/graphtest.go, line 710 at r7](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZGAjzsEfcOqy8fRtLg:-LZGaJ-w6AUKYernBq2p:b-o2lj1s) ([raw file](https://github.com/cayleygraph/cayley/blob/84b79193a6ed422cf624b495c6d5f62bf8fb8d64/graph/graphtest/graphtest.go#L710)):*
<details><summary><i>Previously, fungl164 wrote…</i></summary><blockquote>

I think I see what you meant.  I faked-forced a different hash for Bytes vs IRI by inserting an extra byte.  Dunno if its the best way or if there will be other clashes, but hopefully unlikely if the hash is strong enough... : )
</blockquote></details>

If the extra byte is not enough of a differentiator, adding a stronger suffix and/or prefix should make it work... : )

fungl164:
___
*[quad/value.go, line 66 at r5](https://reviewable.io/reviews/cayleygraph/cayley/766#-LZ0Lp6h2dIra5QIcpZH:-LZGhLIu9S9d4JHaeYu2:b-896fix) ([raw file](https://github.com/cayleygraph/cayley/blob/9c5a6c674b39e9301e500146cef35390b4682dc2/quad/value.go#L66)):*

Let's make it a `\x00` prefix instead of a `\x01` suffix. If later we decide to change the hash for other value, we will use this extra byte as a value type ID.
</blockquote></details>

Done.

iddan:
What is the status of this PR?

dennwc:
Need to rebase and check it again.

PR #767, opened 2019-02-15
kv: get badger to work

1l0:
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/767)


dennwc:
It seems you are using the wrong version of Badger. Can you try building Cayley with `GO111MODULE=on`?

1l0:
You are right. It was my fault and `go mod` messed me up. Thank you for the clarification.

Issue #779, opened 2019-04-23
memstore is not threadsafe


dpavsrtrl:
Tried running memstore example with -race and and got errors.

After submitting PR #780 in an attempt to make memstore threadsafe, discovered that it can be threadsafe but not consistent for readers when a writer is updating it (by deleting, adding).

Now I believe that the issue is not that current memstore is not threadsafe, but there is no concurrent memstore.

iddan:
Do you consider working on consistent concurrent memstore?

dennwc:
Still relevant. Will need to implement it at some point.

dpavsrtrl:
Ok, I'm sorry. Don't have time currently.

PR #780, opened 2019-04-23
make memstore threadsafe

dpavsrtrl:
#779

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/780)

### DIFF
--- a/graph/memstore/all_iterator.go
+++ b/graph/memstore/all_iterator.go
@@ -102,8 +102,10 @@ func (it *AllIterator) Contains(ctx context.Context, v graph.Value) bool {
 	if !ok {
 		return false
 	}
-	p := it.qs.prim[id]
-	if p.ID > it.maxid {
+	it.qs.mu.RLock()
+	defer it.qs.mu.RUnlock()
+	p, ok := it.qs.prim[id]
+	if !ok || p.ID > it.maxid {
 		return false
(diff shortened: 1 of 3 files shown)


dennwc:
Thanks for working on it, but it may not be as straightforward as it is in this PR.

The problem is that iterators should somehow snapshot the current state of the store. In other cases, they may race with other operations like delete.

dpavsrtrl:
No problem. This is just getting foot in the door. I'll try immutable collections.

dpavsrtrl:
After considering the performance impact of snapshotting/persistence, perhaps it would be desirable to have two versions of memstore: current single-threaded and another, which would be concurrent.

dpavsrtrl:
Now I believe memstore to be threadsafe, but it's certainly not consistent for concurrent writers and readers. Snapshots will require more work.

dpavsrtrl:
These last few commits were simply inertial, I now have no intention of making current memstore threadsafe, unless it's useful for someone like this (threadsafe but not read/write concurrency consistent). I'd rather close this PR in favor of submitting another one with a new backend.

iddan:
@dpavsrtrl so do you consider working on transactions for consistent threaded memstore?

iddan:
@dennwc can we close this?

PR #781, opened 2019-04-29
gkvlite backend

dpavsrtrl:
While searching for existing concurrent memstore backends, I stumbled upon gkvlite. It supports both file and in-memory modes. It's not perfect, though.
-  In-memory stores are safe, but rollbacks are not supported, and concurrent reads are dirty.
- File stores support rollbacks, are threadsafe, but concurrent reads are dirty
- On my machine, benchmarks show it to be faster than badger, but my machine doesn't have SSD

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/781)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
 	github.com/badgerodon/peg v0.0.0-20130729175151-9e5f7f4d07ca
 	github.com/boltdb/bolt v1.3.1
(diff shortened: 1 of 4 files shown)


dennwc:
Sorry for a huge delay. It was required to finish a long refactoring on our side.

Can you please apply the same change to [this project](https://github.com/hidal-go/hidalgo) instead? We moved our backend support there. The interfaces are still the same, so shouldn't be a lot of work.

I'm closing it here, hope to see you again in Hidalgo :)

Issue #784, opened 2019-06-26
Read API to accept match parameters


iddan:
`GET /api/v2/read` should accept `subject`, `predicate`, `object` and `context` parameters to only return matching quads. Similar functionality is implemented in the Cayley Gephi stream API and RDF4J statements API.

dennwc:
Merged upstream.

PR #785, opened 2019-06-26
Read API to accept match parameters

iddan:
Implements #784. This is the first time I write in go :)

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/785)


server/http/api_v2.go
```diff
@@ -29,6 +29,7 @@ import (
 
 	"github.com/cayleygraph/cayley/clog"
 	"github.com/cayleygraph/cayley/graph"
+	"github.com/cayleygraph/cayley/internal/gephi"
```
dennwc: This is a strange dependency for this package. It would be better if you copy the implementation of `valuesFromString` from `gephi` package instead.

The `MatchIterator` is also not a problem. I would say that it's worth rewriting it differently, so it can take advantage of more query optimizations:
```
var q shape.Quads
// same for other 3 directions
if len(subjects) != 0 {
  q = append(q, shape.QuadFilter{Dir: quad.Subject, Nodes: shape.Lookup(subjects)})
}
it := shape.BuildIterator(q, qs)
```

server/http/api_v2.go
```diff
@@ -309,7 +310,13 @@ func (api *APIv2) ServeRead(w http.ResponseWriter, r *http.Request) {
 		jsonResponse(w, http.StatusBadRequest, err)
 		return
 	}
-	qr := graph.NewQuadStoreReader(h.QuadStore)
+	sub := gephi.ValuesFromString(r.FormValue("sub"))
```
dennwc: Can you please update `./docs/api/swagger.yml` as well, so it lists new parameters? You can use https://editor.swagger.io/ to validate it.

graph/shape/shape.go
```diff
@@ -1395,3 +1395,28 @@ func (s Save) Optimize(r Optimizer) (Shape, bool) {
 	}
 	return s, opt
 }
+
+type Values struct {
+	Sub   []quad.Value
+	Pred  []quad.Value
+	Obj   []quad.Value
+	Label []quad.Value
+	From  Shape
```
dennwc: `From` is unused.

graph/shape/shape.go
```diff
@@ -1395,3 +1395,28 @@ func (s Save) Optimize(r Optimizer) (Shape, bool) {
...
+}
+
+func (v Values) BuildIterator(qs graph.QuadStore) graph.Iterator {
```
dennwc: It can be a function returning a shape:
```golang
func QuadsWithValues(sub, pred, obj, label []quad.Value) Shape {
```
And it will be a user's responsibility to call `BuildIterator` or use the shape in a more complex query.

iddan: I'm not sure how to do it

dennwc: I mean, the `Quads` type is a shape, so I only propose to remove the new `Values` type, make a function instead of a method (no receiver) and do not call `BuildIterator` at the end, returning `q` directly. Note the signature I proposed though. We want to return `Shape` in general, not `Quads` specifically, so we don't lock ourself down with a specific implementation.

iddan: Sounds correct. I'm just not fully sure how to implement it.

dennwc: No worries, let me hijack the PR then. Will push the changes here, or in case of issues will close the PR and merge directly to master (will preserve your authorship of course).

dennwc:
Was unable to hijack the branch, since it's made from your `master` branch. Merged manually. Thanks again :)

PR #796, opened 2019-07-08
path.Order()

iddan:
- [x] Basic implementation
 - [x] Tests and bug fixes

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/796)


iddan:
Works:
```javascript
g.V().Order().All()
```
```
{
	"result": [
		{
			"id": "smart_person"
		},
		{
			"id": "<alice>"
		},
		{
			"id": "<are>"
		},
		{
			"id": "<bob>"
		},
		{
			"id": "<charlie>"
		},
		{
			"id": "<dani>"
		},
		{
			"id": "<emily>"
		},
		{
			"id": "<follows>"
		},
		{
			"id": "<fred>"
		},
		{
			"id": "<greg>"
		},
		{
			"id": "<predicates>"
		},
		{
			"id": "<smart_graph>"
		},
		{
			"id": "<status>"
		}
	]
}
```

iddan:
```javascript
g.V("<bob>").Out().Order().All()
```
Retruns:
```json
{
	"result": [
		{
			"id": "<fred>"
		}
	]
}
```
Instead of:
{
	"result": [
                {
			"id": "cool_person"
		},
		{
			"id": "<fred>"
		}
	]
}

iddan:
```javascript
g.V("<bob>").Order().All()
```
Returns
```json
{
	"result": null
}
```
Instead of:
```json
{
	"result": [
		{
			"id": "<bob>"
		}
	]
}
```

iddan:
Works:
```javascript
g.V().Tag("entity").Out("<status>").Tag("status").Order().All()
```
```json
{
	"result": [
		{
			"entity": "<dani>",
			"id": "cool_person",
			"status": "cool_person"
		},
		{
			"entity": "<greg>",
			"id": "cool_person",
			"status": "cool_person"
		},
		{
			"entity": "<emily>",
			"id": "smart_person",
			"status": "smart_person"
		},
		{
			"entity": "<greg>",
			"id": "smart_person",
			"status": "smart_person"
		}
	]
}
```

iddan:
I am going to fix all the issues once this feature gets priority in my company's projects. Probably in less than a month.

iddan:
Updated the code to the new Iterator interface. Ready for rereview

iddan:
Done everything except:
> graph/iterator/sort.go, line 75 at r5 (raw file):
> 
> func (it *sortNext) NextPath(ctx context.Context) bool {
>	return false
> I think this is incorrect. NextPath is not implemented on Sort itself (which is correct), but it may be implemented on the subIt. So you should call it in getSortedValues as well (since we don't know if it will be called by the user or not).
Which I don't understand

iddan:
@dennwc I noticed a bug that after calling `order()` `back()` isn't working as expected. Does it has to do with `NextPath()`?

dennwc:
It might be. You can test it with `CAYLEY_DEBUG_OPTIMIZER=true`, and if you see it inside `Quads` and then `NodesFrom` at some point, than it is the case. As a rule of thumb, right now it should always be at the top level.

iddan:
Then I'll check if the bug persists after we fix `NextPath()`

dennwc:
@iddan I rebased the branch and implemented `NextPath` in it. Please test if it solves the issue you mentioned.

Note that ordering may dissappear in some circumstances when using Paths. So changes to the optimizer must be made to ensure that it works in all cases.

iddan:
Looks like the issue I reported was solved.

PR #809, opened 2019-08-14
ensure the nosql stores registration occurs before they are iterated

3pCode:
ensure the nosql stores registration occurs before they are iterated. This issue occurred when using cayley as a library, previously (prior to hidal-go refactoring) an import of http://github.com/cayleygraph/cayley/graph/nosql/mongo, would be enough to ensure that the mongo store was registered, and a call to registry.NewQuadStore("mongo",dppath,options), would succeed.

The issue in the refactoring was the order of the imports was resulting in the nosql class iterating over its registry before the init of the underlying stores had executed which resulted in skipping their registration.

Switched the order of imports to ensure the no sql stores init themselves into the registry before the nosql store iterates its registry into the main registry. Added a comment to the ordering as vscode can flip them if you have organize imports on save enabled. 

Not quite sure why the changes to the go.sum file, regenerated from scratch, and that is how the generation came out.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/809)


dennwc:
Sorry, totally missed the https://github.com/hidal-go/hidalgo/pull/3 somehow, let me merge that first.

PR #810, opened 2019-08-14
pick up latest hidal-go

3pCode:
pick up latest-hidal go, which uses the official mongo driver

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/810)


dennwc:
Can you please rebase on the current `master` instead of merging it? Commits in `master` have a bit different hash because the way how Github rebases things on merge.

3pCode:
will do.

PR #842, opened 2019-09-28
Refactored Dockerfile

iddan:
- Simplified the Dockerfile
 - Added more documentation to it
 - Updated go base image

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/842)


Dockerfile
```diff
@@ -1,66 +1,37 @@
...
 
-# Restore dependencies
+# Install dependencies first
 COPY go.mod go.sum ./
 RUN go mod download
 
-# This will be used to init cayley and as config file in the final image.
-# Make sure you start every path with %PREFIX% to make it available in both 
-# the builder image and the final image.
-RUN echo '{"store":{"backend":"bolt","address":"%PREFIX%/data/cayley.db"}}' > config.json
```
dennwc: I don't think we should change the default configuration for Docker.

First, it's a breaking change.

Second, I think you usually expect a persistence from services running in Docker, so things like:
```
docker stop cayley
docker start cayley
``` 
shouldn't drop your data.

Yes, it's inconsistent with CLI, but CLI use case is different - it's mostly for interactive usage. So you can quickly load any nquads file to memory, play with it, and setup a proper config with persistence if needed.

So instead of changing the default behavour for either CLI or Docker I think we should document it better.

Dockerfile
```diff
@@ -1,66 +1,37 @@
...
+    -installsuffix cgo \
+    -o /fs/bin/cayley \
+    -v \
+    ./cmd/cayley
 
 FROM scratch
-LABEL maintainer="Yannic Bonenberger" \
-           email="contact@yannic-bonenberger.com"
 
-# Expose the port and volume for configuration and data persistence. If you're
```
dennwc: I think this comment and `VOLUME` was quite useful in terms of docs.

Dockerfile
```diff
@@ -1,66 +1,37 @@
...
-# using a backend like bolt, make sure the file is saved to this directory.
 COPY --from=builder /fs /
-VOLUME ["/data"]
 
 EXPOSE 64210
 
-# Adding everything to entrypoint allows us to init+load+serve
```
dennwc: This comment was useful as well :)

Also, please preserve line break before the EOF. Some tools require them.

Dockerfile
```diff
@@ -1,66 +1,37 @@
-FROM golang:1.11 as builder
+FROM golang:1.13.1 as builder
```
dennwc:
```suggestion
FROM golang:1.13 as builder
```
so we always pick the security patches automatically

Dockerfile
iddan: I agree. So I think we should streamline the behaviour the other way in the next breaking change. Similarly to other database persistency should be the default.

iddan:
Ready for re-review @dennwc

Dockerfile
```diff
@@ -1,66 +1,56 @@
...
-RUN mkdir -p /fs/etc/ssl/certs
-RUN cp /etc/ssl/certs/ca-certificates.crt /fs/etc/ssl/certs/ca-certificates.crt
+# Set up workdir for compiling
+WORKDIR /src
 
-RUN mkdir -p /fs/lib/x86_64-linux-gnu
-RUN cp /lib/x86_64-linux-gnu/libc-* /fs/lib/x86_64-linux-gnu/
+# Copy dependencies and install first
+COPY go.mod go.sum ./
+RUN go mod download
```
dennwc: That's a very good idea - will make sequential builds a lot faster.

PR #846, opened 2019-10-04
VSCode Configuration

iddan:
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/846)


dennwc:
Sorry, but I don't think it's a good idea. IDE config shouldn't be in the repository, even if it's a run configuration.

I use a different IDE for example, so pushing all the configs for all the IDEs we use just doesn't make sense.

iddan:
It's usually what I do with projects. But I understand your logic.

Issue #845, opened 2019-10-02
Include assets with binary


iddan:
**Description**
Currently, static assets are included in the release of Cayley separately from the binary. It makes things a little more complex in the Dockerfile and maybe other releases. Instead, we can include the static assets as part of the binary with https://github.com/gobuffalo/packr and reduce the complexity.

iddan:
@dennwc What do you think?

dennwc:
Yes, I think it's a good idea. We can try including it into v0.7.6.

PR #847, opened 2019-10-04
Include assets in binary

iddan:
- Use [packr](https://github.com/gobuffalo/packr) to include static assets inside the binary
 - Refactor serving code for static assets (UI, docs, static)
 - Remove assets directory related code
 - Remove `sync-assets.sh` script

Resolves: https://github.com/cayleygraph/cayley/issues/845

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/847)


Dockerfile
```diff
@@ -1,9 +1,12 @@
 FROM golang:1.13 as builder
 
+# Install packr
+RUN go get -u github.com/gobuffalo/packr/packr
```
dennwc: Can we use [v2](https://github.com/gobuffalo/packr/blob/master/v2/README.md) instead?

Dockerfile
```diff
@@ -21,18 +24,16 @@ ADD . .
 
 # Pass a Git short SHA as build information to be used for displaying version
 RUN SHORT_SHA=$(git rev-parse --short=12 HEAD); \
-    go build \
+    packr build \
```
dennwc: I'm worried about this one a bit. Is it possible to generate files, commit and `go build` without `packr`?
This will prevent GoReleaser from compiling assets to the binary, I believe.

cmd/cayley/command/http.go
```diff
@@ -46,7 +46,6 @@ func NewHttpCmd() *cobra.Command {
 	cmd.Flags().String("host", "127.0.0.1:64210", "host:port to listen on")
 	cmd.Flags().Bool("init", false, "initialize the database before using it")
 	cmd.Flags().DurationP("timeout", "t", 30*time.Second, "elapsed time until an individual query times out")
-	cmd.Flags().StringVar(&chttp.AssetsPath, "assets", "", "explicit path to the HTTP assets")
```
dennwc: Can we preserve this option and set it to Packr somehow? Since paths given to Packr boxes are `../../xxx`, resolving local UI files won't work if they are in the same dir.

docs/Contributing.md
```diff
@@ -12,29 +12,39 @@ export GO111MODULE=on
 
 Follow the instructions for running Cayley locally:
 
-```
+````
```
dennwc: Typo?

cmd/cayley/command/http.go
iddan: When will they be in the same dir? Assuming the assets are provided inside the binary.

Dockerfile
iddan: I'm not sure, it's my first time using packr but I'll check. Why should we prevent GoReleaser from compiling assets to the binary?

dennwc: Because it will call plain `go build`, not `packr build`. We can probably override that, but I would prefer not to, if we can.

cmd/cayley/command/http.go
dennwc: It's probably a good idea to do it in Docker, for example. The point is that we can override the embedded UI for local development, or use UI with a build that has none embedded into it.

iddan: I think that if you want to use custom UI it’s legit it will be served by a different process. For instance: Cayley Web has it’s own dev server

dennwc: Fair enough. Then lets set this flag to a new variable (`.StringVar(&assetsPath, ...`) and fail the CLI with an error if the value of the flag is not empty. This will at least make sure that someone who have used the flag to override the UI gets an error message explaining that the flag is deprecated now.

Dockerfile
iddan: Seems like V2 solves this exact problem, I'll update the code to use it.

cmd/cayley/command/http.go
iddan: Done

PR #853, opened 2019-10-12
fixed the import in example code

alimoeeny:
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/853)

### DIFF
--- a/docs/quickstart-as-lib.md
+++ b/docs/quickstart-as-lib.md
@@ -10,7 +10,7 @@ import (
     "log"
 
     "github.com/cayleygraph/cayley"
-    "github.com/cayleygraph/quad"
+    "github.com/cayleygraph/cayley/quad"
 )
 
 func main() {


docs/quickstart-as-lib.md
```diff
@@ -10,7 +10,7 @@ import (
     "log"
 
     "github.com/cayleygraph/cayley"
-    "github.com/cayleygraph/quad"
+    "github.com/cayleygraph/cayley/quad"
```
iddan: This import is intended to be github.com/cayleygraph/quad, we have a separate repository for it.

alimoeeny: hmm, yes.
Basically if  you copy the example code, and try to run it with `go run .`
under go 1.13, it fails. 
And the proposed change is needed to make it work.

iddan:
What do you think @dennwc

PR #869, opened 2019-10-14
Fix redirect to correct files.

saguywalker:
Fix redirect to correct files in README.md.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/869)


iddan:
Thank you!

PR #876, opened 2019-10-16
Inference

iddan:
This PR implements an in-memory store for the RDFS type system and functions used for inference.
`inference_test` covers 100% of the statements to make sure no edge case is missed.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/876)


iddan:
@dennwc I think it is ready for rereview

PR #881, opened 2019-10-27
LinkedQL Schema

iddan:
This PR implements OWL schema generation for LinkedQL to be used for code generation in client libraries. The only missing part is including comments for docs and testing properly.

This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/881)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/cayleygraph/cayley
 go 1.12
 
 require (
+	github.com/alecthomas/jsonschema v0.0.0-20191017121752-4bb6e3fae4f2 // indirect
 	github.com/badgerodon/peg v0.0.0-20130729175151-9e5f7f4d07ca
 	github.com/cayleygraph/quad v1.1.0
 	github.com/cockroachdb/apd v1.1.0 // indirect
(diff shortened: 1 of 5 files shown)


iddan:
I've created a repository demonstrating how the schema will be used in client libraries: https://github.com/cayleygraph/python-client

iddan:
Currently missing: unify properties (use `owl:unionOf` for domain and assuring range)

iddan:
Also: the ability to easily consume the produced schema file from other repositories.

iddan:
Implemented `owl:unionOf`

iddan:
The vision is: creating a JSON-LD OWL document that can be fetched by client libraries at build time and be used to generate the objects where the steps structure are translated to methods on Path object. The python client example demonstrates that by creating a Path class. This approach will work for JavaScript as well. For Go, it will be similar but with methods for Path struct.
The document specifies the structures, their properties (which are translated to arguments for the methods), and the types those properties accept (e.g. single/multiple values, and the specific type of values).

iddan:
Should use Description for adding a comment property to the schema

PR #888, opened 2019-11-09
External UI & Health Route & HTTP Refactor

### LINKED ISSUE #500: Decouple web interface from database
Having the web interface mixed with the database source code is quite confusing. I'm pretty sure at least 2 directories are for the web interface, `static/` and `templates/`; I don't know/understand if there are other.

Would it be possible to separate all the code for the web interface from the database code, for example in a single sub-directory or on a different repository completely? This should also make it easier to plug in different interfaces at will; for example you can ship cayley with a minimalist one by default, and let people replace this code with another interface if they prefer.

### LINKED ISSUE #804: Add Health Status endpoint
**Description**
When the server starts there is a delay until it listens to incoming requests.

**Steps to reproduce the issue:**
1. Start a server
2. Immediately try to connect 
3. Connection failure

**Expected results:**

Server to receive the request and response when ready.

iddan:
- Remove old UI code
 - Stop serving docs from HTTP server
 - Use UI from http://github.com/cayleygraph/web
 - Refactor `internal/http` and `/server/http/api_v2.go`
 - Add `/health` route which always returns `204`
 - Add CLI command `cayley health`
 - Add `cayley health` as Docker health check
 - FIxed emitted status code

Resolves #500 
Resolves #804

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/888)


iddan:
@dennwc I think it's ready for rereview

PR #896, opened 2019-12-11
Extracted Client

iddan:
Code generated a client package with [OpenAPI Generator](https://github.com/OpenAPITools/openapi-generator) based on `docs/api/swagger.yml`. 

 - [x] Corrected types in `swagger.yml`
 - [x] Refactor HTTP Server tests to not rely on client

This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/896)

PR #897, opened 2019-12-15
LinkedQL Client

iddan:
Code-generate LinkedQL Go client according to the LinkedQL schema.
This PR includes a minimal implementation of an OWL in-memory store to support the code generation.

### Missing
 - [ ] Documentation and example of integration with the [new client library](https://github.com/cayleygraph/cayley/pull/896)
 - [ ] Method to unmarshal received data structures (something like `json.Unmarshal` but for JSON-LD)


iddan:
@dennwc should this be moved somewhere else?

PR #900, opened 2019-12-16
Inference: Domain Range Implicit Instances

iddan:
- Implement RDFS 2 and 3 rules: domain and range implicit class instance
 - Restore 100% test coverage

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/900)

### DIFF
--- a/inference/inference.go
+++ b/inference/inference.go
@@ -19,7 +19,7 @@
 //
 // Exported from: https://www.researchgate.net/figure/RDF-RDFS-entailment-rules_tbl1_268419911
 //
-// Implemented here: 1 5 6 8 10 11
+// Implemented here: 1 2 3 5 6 8 10 11
 package inference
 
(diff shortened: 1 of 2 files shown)


inference/inference.go
```diff
@@ -310,7 +313,15 @@ func (s *Store) processQuad(q quad.Quad) {
 	case rdfs.Range:
 		s.setPropertyRange(sub, obj)
 	default:
-		s.addPropertyInstance(pred)
+		p := s.addPropertyInstance(pred)
+		domain := p.Domain()
+		if domain != nil {
+			domain.references++
+		}
+		_range := p.Range()
```
dennwc: Sorry, I was the one mentioning `_range`, but it would be better to rename it to `prange` (property range) or something similar (and short). Same in other places.

PR #903, opened 2019-12-27
fix links in quickstart-as-application

verygoodsoftwarenotvirus:
Was perusing the Cayley docs and happened upon this link that didn't work. Years of internet spelunking have steeled me for such moments, so I quickly found the docs, but the whole point of open source is so that someone does that sort of work once.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/903)


iddan:
Amazing! Thank you.

PR #904, opened 2019-12-27
LinkedQL: Context step

iddan:
Depends on #902 

 - [x] Define context step
 - [x] Add voc.Namespaces to all steps path building
 - [x] Separate steps definition to easily find where IRIs need to be parsed
 - [x] Removed Type() method to reduce `quad.IRI` references
 - [x] Parse all IRIs using the newly provided ns

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/904)


iddan:
In another PR we are going to make all the names in LinkedQL private as the language should only be consumed

iddan:
"Context" will not be an actual step in the schema. Rather later we'd refactor it to be implemented per client  as it is more of a JSON-LD serialization thing than a logical step.

dennwc:
As agreed, merging earlier, feedback will be addressed in the following PRs.

PR #905, opened 2019-12-28
LinkedQL: Match

iddan:
Depends on #904 
Define a `Match` step in LinkedQL that accepts a subgraph pattern and return entities matching the graph.
Example:
```go
Vertex().Match({ "http://example.org/likes": "http://example.org/bob" })
```
Will return all the entities which have a property pointing to Bob.

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/905)

### DIFF
--- a/internal/linkedql/schema/schema.go
+++ b/internal/linkedql/schema/schema.go
@@ -2,7 +2,9 @@ package schema
 
 import (
 	"encoding/json"
+	"fmt"
 	"reflect"
+	"strconv"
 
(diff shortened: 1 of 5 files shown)


iddan:
Quoting the Zen of Python: "There should be one—and preferably only one—obvious way to do it.". Should we remove steps that can be replaced with `.match()`?
Steps that can be replaced with `.match()`:
 - `.Is()`
 - `.Entity()`
 - `.Entities()`
 - `.Has()`
 - `.HasReverse()`

PR #942, opened 2020-07-07
LinkedQL: Add context for test cases data

iddan:
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/942)


amansx:
@iddan just remove the duplicate files keep your test files the same case

iddan:
I'm not sure how to do that.

PR #954, opened 2021-06-02
memstore add rwmutex, fix multi-thread issue

RyouZhang:
memstore add rwmutex, fix multi-thread issue

### DIFF
--- a/graph/memstore/quadstore.go
+++ b/graph/memstore/quadstore.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/cayleygraph/cayley/graph"
 	"github.com/cayleygraph/cayley/graph/iterator"
@@ -44,7 +45,7 @@ type bnode int64
 func (n bnode) Key() interface{} { return n }
 
 type qprim struct {
-	p *Primitive
+	p *primitive
 }
 
 func (n qprim) Key() interface{} { return n.p.ID }
@@ -60,7 +61,7 @@ type QuadDirectionIndex struct {
 }
 
 func NewQuadDirectionIndex() QuadDirectionIndex {
-	return QuadDirectionIndex{[...]map[int64]*Tree{
+	return QuadDirectionIndex{index: [...]map[int64]*Tree{
 		quad.Subject - 1:   make(map[int64]*Tree),
 		quad.Predicate - 1: make(map[int64]*Tree),
 		quad.Object - 1:    make(map[int64]*Tree),
@@ -72,6 +73,7 @@ func (qdi QuadDirectionIndex) Tree(d quad.Direction, id int64) *Tree {
 	if d < quad.Subject || d > quad.Label {
 		panic("illegal direction")
 	}
+
 	tree, ok := qdi.index[d-1][id]
 	if !ok {
 		tree = TreeNew(cmp)
@@ -88,7 +90,7 @@ func (qdi QuadDirectionIndex) Get(d quad.Direction, id int64) (*Tree, bool) {
 	return tree, ok
 }
 
-type Primitive struct {
+type primitive struct {
 	ID    int64
 	Quad  internalQuad
 	Value quad.Value
@@ -137,12 +139,17 @@ type QuadStore struct {
 	// TODO: string -> quad.Value once Raw -> typed resolution is unnecessary
 	vals    map[string]int64
 	quads   map[internalQuad]int64
-	prim    map[int64]*Primitive
-	all     []*Primitive // might not be sorted by id
+	prim    map[int64]*primitive
+	all     []*primitive // might not be sorted by id
 	reading bool         // someone else might be reading "all" slice - next insert/delete should clone it
 	index   QuadDirectionIndex
 	horizon int64 // used only to assign ids to tx
 	// vip_index map[string]map[int64]map[string]map[int64]*b.Tree
+
+	vrw sync.RWMutex
+	qrw sync.RWMutex
+	prw sync.RWMutex
+	irw sync.RWMutex
 }
 
 // New creates a new in-memory quad store and loads provided quads.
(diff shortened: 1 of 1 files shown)


graph/memstore/quadstore.go
```diff
@@ -178,7 +185,10 @@ func (qs *QuadStore) addPrimitive(p *Primitive) int64 {
 }
 
 func (qs *QuadStore) appendPrimitive(p *Primitive) {
+	qs.prw.Lock()
 	qs.prim[p.ID] = p
+	qs.prw.Unlock()
+
 	if !qs.reading {
 		qs.all = append(qs.all, p)
```
dennwc, line 193: This will cause races as well.

graph/memstore/quadstore.go
```diff
@@ -143,6 +145,11 @@ type QuadStore struct {
 	index   QuadDirectionIndex
 	horizon int64 // used only to assign ids to tx
 	// vip_index map[string]map[int64]map[string]map[int64]*b.Tree
+
+	vrw sync.RWMutex
```
dennwc, line 149: Please use [mutex hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2), so it's clear what these mutexes protect.

graph/memstore/quadstore.go
```diff
@@ -198,24 +208,36 @@ func (qs *QuadStore) resolveVal(v quad.Value, add bool) (int64, bool) {
...
 				return id, ok
 			}
+			qs.prw.RUnlock()
 			qs.appendPrimitive(&Primitive{ID: id, refs: 1})
 			return id, true
 		}
 	}
 	vs := v.String()
+	qs.vrw.RLock()
 	if id, exists := qs.vals[vs]; exists || !add {
```
dennwc, line 226: If the mutexes are that granular, you would probably want to wrap accesses to `qs.vals` (and other things) into separate functions, so you don't have to put Lock/Unlock everywhere.

dennwc (changes requested):
Thanks for the contribution! I appreciate your effort of solving this issue, but you'll need to add tests to prove that it's not racy. A few previous attempts to fix it failed by either racing or deadlocking.

graph/memstore/quadstore.go
RyouZhang: ok, I'll add a multi-thread add quad test case

graph/memstore/quadstore.go
RyouZhang: I'll fix it, thanks

graph/memstore/quadstore_test.go
```diff
@@ -251,3 +256,42 @@ func TestTransaction(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, st, st2, "Appended a new quad in a failed transaction")
 }
+
+// test multi thread insert and query
+func TestMultiThreadQuery(t *testing.T) {
```
dennwc, line 261: Can you please add this test to CI with `-race` flag, so it's clear there's none (and none are introduced in the future)?

Also, any chance to not adding more dependencies? Or using something like `x/sync/errgroup`. No offense, it's just more widely used, and there's a chance we already have it as a transitive dependency.

graph/memstore/quadstore_test.go
```diff
@@ -251,3 +256,42 @@ func TestTransaction(t *testing.T) {
...
+				followers, err := path.StartPath(qs, quad.Raw("G")).In("follows").Iterate(ctx).AllValues(qs)
+				if err != nil {
+					return nil, err
+				}
+				return followers, nil
+			}
+		}
+	}
+
+	results := async.All(funcs, 1*time.Second)
```
dennwc, line 290: Would be also nice to test results for consistency somehow, but that might not be trivial. Could be done later.

graph/memstore/quadstore.go
```diff
@@ -143,6 +145,11 @@ type QuadStore struct {
 	index   QuadDirectionIndex
 	horizon int64 // used only to assign ids to tx
 	// vip_index map[string]map[int64]map[string]map[int64]*b.Tree
+
+	valsMu  sync.RWMutex
```
dennwc, line 149: Would you mind applying the "mutex hat" as I [mentioned earlier](https://github.com/cayleygraph/cayley/pull/954#discussion_r643769984)?

For example, if to access `vals` you only need `valsMu`, then place then together (with mutex on top), separated from other variables with a line break:
```go
// ...

valsMu sync.RWMutex
vals map[string]int64

// ...
```

or group with anonymous struct:

```go
vals struct{
   sync.RWMutex
   m map[string]int64
}
```

RyouZhang: I don't it clean for code

dennwc (changes requested):
Sorry, but I still cannot accept your change :(

1. You should still add a separate test run to the CI with a `-race` flag. Without it I have no idea if the change achieves what it's supposed to.

2. It's still completely unclear for me how mutexes relate to protected data. For example,  it's clear that `valsMu` is related to `vals`, but is it safe to hold it while you hold `indexMu`? Or should one hold both while writing, or only one? Is it safe for iterator to read primitives and index, while it's modified? I don't see anything protecting `*Tree` objects, for example. You should use "index hat" as I proposed, or clearly explain what can and cannot be done with those fields.

3. It seems to me there quite a few holes in the current implementation, because locks just surround related fields. Usually it only masks the problem with race condition instead of fixing it. I assume currently it's possible to get multiple duplicate nodes if the code happen to write the same node concurrently, probably the same will happen to quads. Your test doesn't show that the data is consistent, only the fact that it won't crash (hopefully).

RyouZhang:
> Sorry, but I still cannot accept your change :(
> 
> 1. You should still add a separate test run to the CI with a `-race` flag. Without it I have no idea if the change achieves what it's supposed to.
> 2. It's still completely unclear for me how mutexes relate to protected data. For example,  it's clear that `valsMu` is related to `vals`, but is it safe to hold it while you hold `indexMu`? Or should one hold both while writing, or only one? Is it safe for iterator to read primitives and index, while it's modified? I don't see anything protecting `*Tree` objects, for example. You should use "index hat" as I proposed, or clearly explain what can and cannot be done with those fields.
> 3. It seems to me there quite a few holes in the current implementation, because locks just surround related fields. Usually it only masks the problem with race condition instead of fixing it. I assume currently it's possible to get multiple duplicate nodes if the code happen to write the same node concurrently, probably the same will happen to quads. Your test doesn't show that the data is consistent, only the fact that it won't crash (hopefully).

You are right, the patch make it can't crash; if you write concurrently, the data isn't consistent. But use write queue, you can get data concurrently. 
I think memory store is a special case, only for query speed, so if you want to write/read concurrently, you must make write single thread, read concurrently.

CHneger (approved):

CHneger (changes requested):
I need these changes urgently

SVyatoslavG:
What needs to be done to merge these changes? I am experiencing panics due to this bug.

PR #959, opened 2021-06-17
refactor: return errors from ValueOf, NameOf, QuadDirection

paralin:
Adds error return values to ValueOf(), NameOf(), QuadDirection on the QuadStore. The QuadStore currently returns a value for these functions without an error, but internally the QuadStore might run into some errors when processing the value. This hack is clearly identified in the code: rather than return the error, it's just logged.

If the QuadStore has an error while processing ValueOf or NameOf, the returned value will be nil which might lead to many types of undefined behavior - such as queries returning "nil" for one field of a Quad.

Fixes #985

---
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/959)


graph/gaedatastore/iterator.go
```diff
@@ -171,7 +176,12 @@ func newIteratorNext(qs *QuadStore, k string, d quad.Direction, t *Token) *itera
 		clog.Errorf("Cannot create iterator without a valid context")
 		return &iteratorNext{done: true}
 	}
-	name := quad.StringOf(qs.NameOf(t))
+	tn, err := qs.NameOf(t)
+	if err != nil {
+		clog.Errorf("Creating iterator token lookup error: %v", err)
+		return &iteratorNext{done: true}
```
dennwc, line 182:
```suggestion
		return &iteratorNext{done: true, err: err}
```

graph/hasa.go
```diff
@@ -244,10 +253,18 @@ func (it *hasAContains) nextContains(ctx context.Context) bool {
 	for it.results.Next(ctx) {
 		link := it.results.Result()
 		if clog.V(4) {
-			clog.Infof("Quad is %v", it.qs.Quad(link))
+			qlv, err := it.qs.Quad(link)
```
dennwc, line 256: It ignores error here

graph/kv/quadstore.go
```diff
@@ -364,21 +363,19 @@ func (qs *QuadStore) RefsOf(ctx context.Context, nodes []quad.Value) ([]graph.Re
...
+	return vals[0], nil
 }
 
-func (qs *QuadStore) Quad(k graph.Ref) quad.Quad {
+func (qs *QuadStore) Quad(k graph.Ref) (quad.Quad, error) {
 	key, ok := k.(*proto.Primitive)
 	if !ok {
-		clog.Errorf("passed value was not a quad primitive: %T", k)
-		return quad.Quad{}
+		return quad.Quad{}, errors.Errorf("passed value was not a quad primitive: %T", k)
```
dennwc, line 378:
```suggestion
		return quad.Quad{}, fmt.Errorf("passed value was not a quad primitive: %T", k)
```

graph/kv/quadstore.go
```diff
@@ -387,13 +384,13 @@ func (qs *QuadStore) Quad(k graph.Ref) quad.Quad {
...
 	})
+	if err == kv.ErrNotFound {
+		err = nil
+	}
 	if err != nil {
-		if err != kv.ErrNotFound {
-			clog.Errorf("error fetching quad %#v: %s", key, err)
-		}
-		return quad.Quad{}
+		err = errors.Errorf("error fetching quad %#v: %s", key, err)
```
dennwc, line 391:
```suggestion
		err = fmt.Errorf("error fetching quad %#v: %s", key, err)
```

graph/sql/quadstore.go
```diff
@@ -670,9 +675,8 @@ func (qs *QuadStore) NameOf(v graph.Ref) quad.Value {
 		&vtime,
 	); err != nil {
 		if err != sql.ErrNoRows {
-			clog.Errorf("Couldn't execute value lookup: %v", err)
+			return nil, errors.Wrap(err, "error executing value lookup")
```
dennwc, line 678:
```suggestion
			return nil, fmt.Errorf("error executing value lookup: %w", err)
```

graph/sql/quadstore.go
```diff
@@ -624,13 +632,10 @@ func (qs *QuadStore) NameOf(v graph.Ref) quad.Value {
 		panic(fmt.Errorf("unexpected token: %T", v))
```
dennwc, line 632: We could also return an error here now.

internal/gephi/stream.go
```diff
@@ -223,7 +223,11 @@ func (s *GraphStreamHandler) serveRawQuads(ctx context.Context, gs *GraphStream,
 		if qv == nil {
 			continue
 		}
-		q := s.QS.Quad(qv)
+		q, err := s.QS.Quad(qv)
+		if err != nil {
+			// TODO: no error handling
```
dennwc, line 228: Log at least?

query/sexp/parser.go
```diff
@@ -190,7 +190,7 @@ func getIdentString(tree *peg.ExpressionTree) string {
 		}
 	} else {
 		if tree.Value != '"' {
-			out += string(tree.Value)
+			out += string([]rune{rune(tree.Value)})
```
dennwc, line 193: I think converting `->rune->string` directly (without a slice) would work as well.

paralin: That throws an error when you try to build - saying that string(rune) does not have the intended behavior in most cases. Actually makes "go test" fail, so that was why I changed it here.

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/peterh/liner v0.0.0-20170317030525-88609521dc4b
 	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
 	github.com/piprate/json-gold v0.3.0
+	github.com/pkg/errors v0.8.1
```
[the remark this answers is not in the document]
paralin: for errors.Wrap in some cases it's useful to be able to Unwrap the error and get the underlying error

... but I can change them all to be fmt.Errorf("error executing value lookup: %v", err) if needed

graph/hasa.go
paralin: Previously if there was an error it would print "Quad is nil"

graph/gaedatastore/iterator.go
paralin: Are you sure? Given that there are a lot of other locations in the file where errors return done: true without the error field?

graph/hasa.go
paralin: added logging

paralin:
@dennwc thanks for review, changes applied.

graph/gaedatastore/iterator.go
dennwc: I think that might be a copy-paste error or something. Or it might not check for NotFound errors properly.

dennwc (approved):

paralin:
I haven't had a chance to check the docs thoroughly, might want to make an issue to re-check that later.

Issue #971, opened 2022-10-15
Repl Allows One Command to Execute and Subsequent Commands Fail


joshglenn:
**Description**

When I use the repl, I can run one command and that's it. After that I get an error.

**Steps to reproduce the issue:**
1. run repl
2. run 2 + 2 (works)
3. run 2 + 1 (error)

**Received results:**

```
cayley> 2 + 2

=> 4
-----------
1 Result
Elapsed time: 1.95 ms

cayley> 2+1

Error:  iterator closed at <eval>:1:1(0)
cayley> 
```

**Expected results:**

It should allow me to run multiple commands

**Output of `cayley version` or commit hash:**

```
ERROR: logging before flag.Parse: E1014 21:20:53.997357   97410 cayley.go:142] The assets flag is deprecated as assets are provided as part of the binary.
Cayley version: 0.7.7
Git commit hash: dcf764f
```

**Environment details:**

Backend database: `bolt` version unknown

catundercar:
I found the problem:

1. When run a command, the session of `gizmo` will new a iterator. After the first command done, the iterator will be closed.
2. The problem occurs at the time of closing iterator：
![image](https://user-images.githubusercontent.com/24569795/211765387-2f558d85-71a5-46e4-9639-9671474035a2.png)
![image](https://user-images.githubusercontent.com/24569795/211765446-5a027b99-748b-45b2-8f34-499d01acb55a.png)
It will make the `vm` be interrupted.
3.Then run the second command, the `vm` will check `interrupted`.
![image](https://user-images.githubusercontent.com/24569795/211766159-ca1574ec-8c5d-4111-a333-d784ab466ead.png)
So, panic.

Temporary Solutions：
https://github.com/cayleygraph/cayley/blob/94f1b49053867c5df4e89bad894f474806182e38/query/gizmo/gizmo.go#L326

![image](https://user-images.githubusercontent.com/24569795/211766953-03c77511-fd71-49dc-b38c-8dd204ac4cb8.png)

PR #974, opened 2023-01-11
fix: Repl with `gizmo` close iterator interrupted

catundercar:
Related issue: https://github.com/cayleygraph/cayley/issues/971


jmonroynieto:
Compiling it myself for now. I really do hope Cayley becomes active again.

PR #979, opened 2023-05-26
fix: check inner quad iterator for errors when closing

paralin:
This caused a bug where Scan returned an error which was ignored by the iterator. The inner iterator it.Err() would have returned the error correctly, but it was ignored in QuadIterator.

Fixed by checking the inner iterator error when closing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/cayleygraph/cayley/979)

### DIFF
--- a/graph/kv/quad_iterator.go
+++ b/graph/kv/quad_iterator.go
@@ -123,6 +123,9 @@ func (it *quadIteratorNext) TagResults(dst map[string]graph.Ref) {}
 
 func (it *quadIteratorNext) Close() error {
 	if it.it != nil {
+		if err := it.it.Err(); err != nil && it.err == nil {
+			it.err = err
+		}
 		if err := it.it.Close(); err != nil && it.err == nil {
 			it.err = err
 		}
@@ -180,7 +183,8 @@ func (it *quadIteratorNext) Next(ctx context.Context) bool {
 				it.ids = nil
 				it.buf = nil
 				if !it.it.Next(ctx) {
-					it.Close()
+					// note: it.Err() is checked in Close()
+					_ = it.Close()
 					it.done = true
 					return false
 				}


paralin:
@dennwc If you'd like, I'd be happy to take on the role of merging minor fixes & updating dependencies (maintenance) in Cayley's main repository. I have a more "destructive" [fork](https://github.com/aperturerobotics/cayley) which takes some more significant changes but would only apply minor fixes / improvements here.

paralin:
@dennwc this was the bug that, coupled with the liberal use of context.TODO in the codebase, caused a few very hard to trace issues to occur some percentage of the time.

graph/kv/quad_iterator.go
```diff
@@ -123,6 +123,9 @@ func (it *quadIteratorNext) TagResults(dst map[string]graph.Ref) {}
 
 func (it *quadIteratorNext) Close() error {
 	if it.it != nil {
+		if err := it.it.Err(); err != nil && it.err == nil {
```
dennwc, line 126: This is not the right solution, I'm afraid. You'll need to find the "bad" iterator that doesn't propagate its error in `Close`. The code below already checks for that specifically.

And actually, the "bad" iterator is doing something worse: it's not recording `it.err` on `Next/Contains/NextPath`.  Otherwise usual iterator loop would catch it from `it.Err()`, instead of waiting for `Close` to return it. Usually we do not expect `Close` error to be handled, since it's usually too late.

If you would ask me to guess, I'll say the cause is probably somewhere in one of the `Contains` implementations.

dennwc (changes requested):

graph/kv/quad_iterator.go
paralin: That may be so, but it's also valid to check Err the way it's done here, and foolproof, yes?

dennwc: Valid, but not foolproof, since it's just one iterator that does it. For it to be foolproof you'll need to go over ALL iterators and do the same change. It might be easier to locate a "bad" iterator and/or write a test or static check that does.

paralin: I'll continue searching for the other bad iterators.

That said; it seems logical to check Err() here and set it if not nil, especially since the following two Close calls will overwrite it.err if set.

This is a good fix for the bug in the short term imo - thoughts?

dennwc: That's the thing: why `it.it.Close()` below doesn't return an error in the first place? This is the bug worth fixing.

If you have a case that you can debug, just do this (temporarily):

```go
err2 := it.it.Err()
if err := it.it.Close(); err != nil && it.err == nil {
	it.err = err
} else if err == nil && err2 != nil {
	panic(fmt.Errorf("incorrect implementation of %T.Close: Err must return same error as Close", it.it))
}
```

Once you trigger it, you know which `Close` needs to be fixed.

